PR #5, opened 2026-04-09
[META]: Add Project Metadata Preceding Project Files

spencrr:
## Description
Adds python project infra for project before importing project files, specifically:

- gitignore, etc.
- uv/python project files
- pre-commit


.gitignore
```diff
@@ -0,0 +1,381 @@
+# Created by https://www.toptal.com/developers/gitignore/api/linux,macos,windows,python,visualstudiocode,pycharm
```
nina-msft, line 1: does this file need all of this? cross-referencing pyrit's .gitignore and i think we just include python & vscode & pycharm ++ pyrit specific configs and some precommit / docker things

(although if this is standard maybe we want to make a work item to update pyrit's gitignore to snap to this to we can better support the operating systems)

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+    "azure-kusto-ingest>=6.0.3",
+    "jsonschema>=4.26.0",
+    "msgraph-sdk>=1.55.0",
+    "playwright>=1.58.0",
+    "pyrit",
+    "pytest>=9.0.3",
+    "pytest-asyncio>=1.3.0",
+    "pyyaml>=6.0.3",
+]
+
```
nina-msft, line 24: maybe missing

```
[tool.setuptools.packages.find]
where = ["."]
include = ["rampart*"]

[tool.setuptools.package-data]
"*" = ["*.txt", "*.md"]
```

pyproject.toml
```diff
@@ -0,0 +1,53 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "RAMPART"
+version = "0.1.0"
```
nina-msft, line 7: sanity check: this is bumped up to `0.1.1.dev0` in ado but i think we can make this change after our first gh release, so leave for now?

nina-msft (approved):
lgtm pending light comments :)

pyproject.toml
spencrr: 1. https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#custom-discovery -> we are doing default, so not required
2. there are not any txt or md so we can omit for now. If we really want to include the LICENSE etc., I suppose we can add

.gitignore
nina-msft: also left a comment on #6 that we should cross-check with rampart main's `.gitignore` file (i didn't notice we had one already)- so if its easier to do that in #6 vs this pr feel free to close this comment

pyproject.toml
spencrr: yes it is hard to sync two trees

.gitignore
spencrr: this is my go-to maybe good for PyRIT too, yeah! I saw there were some (likely) unnecessary things in ADO so just adding base for now and we add back project-specific ones once they are solidified

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+license = "MIT"
+requires-python = ">=3.11"
+dependencies = [
+    "azure-identity>=1.25.3",
+    "azure-kusto-data>=6.0.3",
+    "azure-kusto-ingest>=6.0.3",
+    "jsonschema>=4.26.0",
+    "msgraph-sdk>=1.55.0",
+    "playwright>=1.58.0",
+    "pyrit",
```
nina-msft, line 19: sanity check: assuming no version here bc we want latest?

spencrr: uv.lock has exact pins

pyproject.toml
spencrr: 2. also, I was cross-referencing with PyRIT [pyproject.toml](https://github.com/microsoft/PyRIT/blob/main/pyproject.toml)

PR #13, opened 2026-04-15
[FEAT] Add InjectionHandle `wait_until_ready` hook

nina-msft:
Today, the only mechanism to wait for injected content to become visible to the target agent is a fixed `asyncio.sleep()` driven by `InjectionHandle.indexing_delay_seconds` (see `XPIAExecution._activate_handles_async()`). This works for simple time-based delays but doesn't generalize to scenarios where readiness can be actively verified, such as:

- Polling an API until an injected document appears in search results
- Querying a database row until it becomes visible
- Waiting for a webhook callback or pipeline completion

This PR introduces a `wait_until_ready` async method on the `InjectionHandle` protocol that surfaces can override to implement custom readiness logic. The current sleep-based delay becomes one default implementation of this hook.

---

### Protocol Changes (Breaking)

- **Removed** `indexing_delay_seconds` property from `InjectionHandle` protocol.
- **Added** `wait_until_ready()` async method to `InjectionHandle` protocol — surfaces now control their own readiness strategy instead of exposing a passive delay value.

### New

- **`sleep_until_ready(delay)`** — free helper function in `rampart.core.injection` for surfaces that only need a simple sleep-based readiness wait.
- **`_OneDriveInjection.wait_until_ready()`** — delegates to `sleep_until_ready` using the surface's `indexing_delay`. Future versions will poll the Graph API instead.
- **`TestOneDriveInjectionWaitUntilReady`** — test class verifying the OneDrive wiring to `sleep_until_ready`.

### Changed

- **`XPIAExecution._activate_handles_async`** — replaced sequential `max(indexing_delay_seconds)` + `asyncio.sleep` with concurrent per-handle readiness via `asyncio.TaskGroup`.
- **`OneDriveSurface`** — internal references changed from private attributes (`self._drive_id`, `self._folder_path`) to public property accessors (`self.drive_id`, `self.folder_path`).
- **test_xpia.py** — `_mock_handle` no longer sets `delay` / `indexing_delay_seconds`; cleanup tests now assert `wait_until_ready` was called.
- **test_protocols.py** — structural subtyping tests updated to match the new protocol shape.
- **test_onedrive.py** — assertions use public properties; removed `test_indexing_delay_from_surface`.

### Housekeeping

- Fixed unnecessary f-string prefixes on strings without interpolation in onedrive.py.
- Removed `!.vscode/settings.json` from .gitignore.

---
* unit tests pass
* `pre-commit run --all-files` run w/ success


.gitignore
```diff
@@ -23,7 +23,8 @@
 .LSOverride
 
 # Icon must end with two \r
-Icon
```
nina-msft, line 26: agh, not intentional change. need to restore this line

rampart/surfaces/onedrive.py
```diff
@@ -61,12 +63,14 @@ def __init__(
         drive_id: str,
         folder_path: str,
         indexing_delay: float = DEFAULT_INDEXING_DELAY,
+        readiness_timeout: float = DEFAULT_READINESS_TIMEOUT,
     ) -> None:
         """Initialize with Graph client and OneDrive location."""
         self._graph_client = graph_client
-        self._drive_id = drive_id
```
nina-msft, line 67: will reverse this

docs/work-items/exception-group-infra-error.md
```diff
@@ -0,0 +1,35 @@
...
+The chosen approach should preserve the existing `ON_ERROR` event dispatch so handlers are still notified, while ensuring the test run is never crashed by a transient failure from a surface, adapter, or readiness check.
+
+## Acceptance Criteria
+
+- [ ] `BaseExecution.execute_async` produces `Result(status=SafetyStatus.ERROR)` for exceptions beyond `InfrastructureError`, including `ExceptionGroup` and `TimeoutError`.
+- [ ] When any handle's `wait_until_ready()` raises `TimeoutError` (wrapped in `ExceptionGroup` by `TaskGroup`), the test produces a clean error result — not an unhandled exception.
+- [ ] `ON_ERROR` event handlers are still notified when a non-`InfrastructureError` exception is caught and converted to an error result.
+- [ ] Individual execution strategies (e.g., `XPIAExecution`) do not need to catch and translate exceptions — the base class handles it as a cross-cutting concern.
+- [ ] Unit tests in `test_execution.py` verify the broadened error handling for `ExceptionGroup`, `TimeoutError`, and other non-`InfrastructureError` exceptions.
+- [ ] Existing tests continue to pass (no regression).
```
bashirpartovi: I think this should be removed from here, it is tracked on the board.

nina-msft: accidentally pushed this :'-) its gone now!

bashirpartovi (approved):
Looks great

PR #18, opened 2026-04-20
[MAINT]: Pin PyRIT + tool

spencrr:
## Description

Add `bump-pyrit.sh` — a one-command helper that resolves a PyRIT version tag to its commit SHA and runs `uv add` to update dependencies, `[tool.uv.sources]`, and `uv.lock` in one shot.

Also bumps PyRIT to [0.13.0](https://github.com/microsoft/PyRIT/commits/releases/v0.13.0) and ran `uv sync --upgrade`.

## Breaking changes

None

## Checklist

- [ ] Documentation updated


pyproject.toml
```diff
@@ -113,4 +113,4 @@ known-first-party = ["rampart"]
 convention = "google"
 
 [tool.uv.sources]
-pyrit = { git = "https://github.com/microsoft/PyRIT", rev = "3873c7762c42febcf844c468e1eb4adee6cf2e30" } # v0.12.1
+pyrit = { git = "https://github.com/microsoft/PyRIT", rev = "6dc8b94139757390286bbce7d53c1f7e58e66e29" }
```
nina-msft: nit: add `# v0.13.1` as a comment similar to prev?

scripts/bump-pyrit.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
nina-msft: This is a useful script to have! I'm wondering about the file structure we'd have by adding a `scripts` folder.

If these will just be dev/maintainer helper scripts, I think a more descriptive folder name is necessary. `dev_scripts/bump_pyrit_version.py` maybe?

Also suggest we rename script to `bump_pyrit_version.py` -- even if we don't add the `_version` we should use `_` separators instead of `-`.

nina-msft (changes requested):

pyproject.toml
spencrr: left out originally to avoid nasty regex - readding. Strictly not necessary (version derived from `project.dependencies`) and could become outdated, but adding back for clarity

scripts/bump-pyrit.sh
spencrr: Asked copilot what other repos do - I think scripts actually makes sense for us too:

| Repo | Directory | Contents |
|------|-----------|----------|
| [**uv** `scripts/`](https://github.com/astral-sh/uv/tree/main/scripts) | scripts | Version bumps (`bump-workspace-crate-versions.py`), release scripts, benchmarks, CI helpers, vendor scripts |
| [**ruff** `scripts/`](https://github.com/astral-sh/ruff/tree/main/scripts) | scripts | Rule codegen (`add_rule.py`), release scripts, ecosystem checks, schema updates, conformance tests |
| [**PyTorch** `scripts/`](https://github.com/pytorch/pytorch/tree/main/scripts) | scripts | Build helpers, linters, release automation, hook setup |

scripts/bump_pyrit_version.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+#
+# Bump the pinned PyRIT dependency to a given version tag.
+# Usage: ./scripts/bump_pyrit_version.sh v0.13.0
```
nina-msft, line 6: did you want usage to include the v? previous version didn't have it but assuming this might've been a ctrl+f replace thing :)

nina-msft (approved):

scripts/bump_pyrit_version.sh
spencrr: yep! should include the v - then the 
```
VERSION = 0.13.0
TAG = v0.13.0
```
verison used by PyPI/uv, tag is the git ref 😄

PR #21, opened 2026-04-21
[META]: Fixup coding standards

spencrr:
## Description

Ensures that the coding standard examples are following some of the other rules (namely, using `list` over `typing.List`, etc.) and adding exceptions for dunder methods for total argument count rule. 

## Breaking changes

None

## Checklist

- [ ] Documentation updated


nina-msft (approved):
Great catch! 🚀

PR #28, opened 2026-04-22
[META]: Correct PR template

spencrr:
## Description

Move the PR template - UI seems to not pickup options from the `.github/PULL_REQUEST_TEMPLATE` folder - since we have just the one template, avoiding complexity and moving to `.github`.

## Breaking changes

None

## Checklist

- [x] Documentation updated


nina-msft (approved):
Looked up why this might be the case ([docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository#adding-a-pull-request-template)) -- seems like if we want to use a subdirectory it needs to be within root or `docs/`, or we need to use a query parameter?

> To create multiple pull request templates and use the template query parameter to specify a template to fill the pull request body, type .github/PULL_REQUEST_TEMPLATE/, then the name of your pull request template. For example, .github/PULL_REQUEST_TEMPLATE/pull_request_template.md. You can also store multiple pull request templates in a PULL_REQUEST_TEMPLATE subdirectory within the root or docs/ directories. For more information, see [Using query parameters to create a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/using-query-parameters-to-create-a-pull-request).

Either way, this change works for a single PR template 🫡

spencrr:
> Looked up why this might be the case ([docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository#adding-a-pull-request-template)) -- seems like if we want to use a subdirectory it needs to be within root or `docs/`, or we need to use a query parameter?
> 
> > To create multiple pull request templates and use the template query parameter to specify a template to fill the pull request body, type .github/PULL_REQUEST_TEMPLATE/, then the name of your pull request template. For example, .github/PULL_REQUEST_TEMPLATE/pull_request_template.md. You can also store multiple pull request templates in a PULL_REQUEST_TEMPLATE subdirectory within the root or docs/ directories. For more information, see [Using query parameters to create a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/using-query-parameters-to-create-a-pull-request).
> 
> Either way, this change works for a single PR template 🫡

I was shooting for the UI to show something like this which is supported with the `.github/PULL_REQUEST_TEMPLATE`
but maybe you need the default to be just in `.github` 🤷‍♂️ we only have one for now 🤣 
image

PR #32, opened 2026-04-26
[FEAT]: LLM-based Prompt Driver

bashirpartovi:
## Summary

Adds `LLMDriver`, a new `PromptDriver` implementation that uses a driving LLM to generate the next prompt on each turn based on the agent's prior response and the evaluator's feedback. Complements the existing `StaticDriver` (fixed prompt sequence) for tests that need adaptive behaviour, multi-turn jailbreak attempts, exploratory probes, or attacks that adjust strategy based on what the agent does.

Built on PyRIT's `PromptChatTarget` and `PromptNormalizer`.

---

Also moved `_pyrit`, which was originally a hidden boundary for everything PyRIT-related, to an actual `pyrit_bridge` package. This is to be more open about what RAMPART is built upon — we will always depend on PyRIT.


bashirpartovi:
closing this request, too many conflicts

PR #33, opened 2026-04-26
[FEAT]: LLM-based Prompt Driver

bashirpartovi:
## Summary

Adds `LLMDriver`, a new `PromptDriver` implementation that uses a driving LLM to generate the next prompt on each turn based on the agent's prior response and the evaluator's feedback. Complements the existing `StaticDriver` (fixed prompt sequence) for tests that need adaptive behaviour, multi-turn jailbreak attempts, exploratory probes, or attacks that adjust strategy based on what the agent does.

Built on PyRIT's `PromptChatTarget` and `PromptNormalizer`.

---

Also moved `_pyrit`, which was originally a hidden boundary for everything PyRIT-related, to an actual `pyrit_bridge` package. This is to be more open about what RAMPART is built upon — we will always depend on PyRIT.


.github/workflows/coverage.yml
```diff
@@ -40,15 +40,18 @@ jobs:
...
         id: tests
-        # --cov-fail-under=0 overrides pyproject.toml [tool.coverage.report] fail_under
-        # so pytest doesn't exit non-zero on low coverage; threshold is checked separately below.
-        run: uv run pytest tests/unit --cov=rampart --cov-report=term-missing --cov-fail-under=0
+        # Use 'coverage run' instead of 'pytest --cov' so that coverage
+        # starts BEFORE pytest loads plugins. rampart registers a pytest
+        # plugin via the pytest11 entry point — if coverage starts after
+        # plugin loading, all module-level code imported during plugin
+        # setup is invisible to the tracer.
+        run: uv run coverage run -m pytest tests/unit -q
```
spencrr, line 48: perfect! thanks for fixing this!

.github/workflows/coverage.yml
```diff
@@ -40,15 +40,18 @@ jobs:
...
 
       - name: Coverage summary
         if: ${{ steps.tests.outcome == 'success' }}
         run: |
           echo '## Coverage Report' >> $GITHUB_STEP_SUMMARY
-          uv run coverage report --format=markdown >> $GITHUB_STEP_SUMMARY
+          uv run coverage report --format=markdown --fail-under=0 >> $GITHUB_STEP_SUMMARY
```
spencrr, line 54: as far as failing coverage, wdyt about leaving as-is? I split coverage from lint+test so that we actually _can_ see coverage regressions?

rampart/drivers/llm.py
```diff
@@ -0,0 +1,353 @@
...
+import logging
+import uuid
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import yaml
+from jinja2 import Template
+from pyrit.exceptions import EmptyResponseException
+from pyrit.memory import CentralMemory
+from pyrit.prompt_normalizer import PromptNormalizer
```
nina-msft, line 39: These imports break the PyRIT isolation rule: https://github.com/microsoft/RAMPART/blob/becbaa8a8f4f80553ce0b4b1f4e243ad63aba6b9/.github/instructions/coding-standards.instructions.md?plain=1#L557-L561

Either we update the instructions to include an exception for `rampart/drivers` similar to how we did for `rampart/converters` (since this is a wrapper of `PromptChatTarget`) or we move this logic into helper functions in the `llm_bridge`

nina-msft: Either way this section in the `.github/coding-standards.instructions.md` should be updated now that `_pyrit` is moved to `pyrit_bridge`

rampart/drivers/llm.py
```diff
@@ -0,0 +1,353 @@
...
+        else:
+            # LLMConfig path: create everything from scratch
+            if self._llm is None:
+                msg = (
+                    "LLMDriver: no LLM config and no target — use "
+                    "from_target() or provide an LLMConfig."
+                )
+                raise DriverError(msg)
+            self._target = create_prompt_target(self._llm)
+            self._normalizer = PromptNormalizer()
```
richlundeen, line 182: I think it is probably a good idea to import and use pyrit directly. But I think we may need to update coding standards in `.github/instructions/coding-standards.instructions.md`

rampart/drivers/llm.py
```diff
@@ -0,0 +1,353 @@
...
+            llm: LLM configuration. Required unless ``target`` is provided.
+            target: Pre-configured PromptChatTarget. Mutually exclusive
+                with ``llm`` — provide one or the other.
+            objective: Per-test goal as a natural-language string.
+            injections: Payloads placed in the agent's data sources.
+
+        Raises:
+            TypeError: If both ``llm`` and ``target`` are provided.
+        """
+        if llm is not None and target is not None:
```
richlundeen, line 115: Should we have more validation here? E.g. raise if neither are provided?

richlundeen (approved):

rampart/core/execution.py
```diff
@@ -236,20 +239,22 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
 
         try:
             result = await self._execute_async(adapter=adapter)
-        except InfrastructureError as exc:
+        except (InfrastructureError, DriverError) as exc:
```
nina-msft: I'm glad `DriverError` is included - but this still just exacerbates the issue we see in work item [#8416](https://dev.azure.com/atml/AI%20Red%20Team/_workitems/edit/8416). Can we widen this to catch all exceptions and log their error type as a Result?

Then we can get rid of generic handling on L254-267 and close out the item and make sure all error types are included.

rampart/drivers/llm.py
nina-msft: More details on suggestion if we move the logic:

```
Suggested fix: Move PyRIT-dependent logic into rampart/pyrit_bridge/llm_bridge.py:

 1. Wrap EmptyResponseException handling inside send_user_turn_async (re-raise as DriverError)
 2. Move the CentralMemory.get_conversation() call into a bridge helper (e.g. count_user_turns_async)
 3. Have the bridge construct PromptNormalizer internally

The driver already uses the bridge for create_prompt_target and send_user_turn_async — extending that pattern would fully isolate PyRIT.
```

rampart/drivers/llm.py
spencrr, line 1: some notes from Copilot:

🔴 Must Fix: `LLMDriver` breaks the PyRIT boundary and leaks PyRIT into the public API

> `LLMDriver` imports `pyrit.exceptions.EmptyResponseException`, `CentralMemory`, `PromptNormalizer`, and `PromptChatTarget` directly, then it is re-exported from both `rampart.drivers` and top-level rampart. That means a normal `import rampart` now pulls a PyRIT-backed module into the public surface.
> 
> This conflicts with the repository rule that PyRIT interaction must be isolated behind the bridge (coding-standards.instructions.md) and with the boundary test’s stated intent that public RAMPART modules must not import PyRIT directly (test_llm_bridge.py). The test simply misses `rampart.drivers`, `rampart.drivers.llm`, and `rampart.probes`.
> 
> **Recommendation:** move the PyRIT-owned state into `rampart.pyrit_bridge`. For example, introduce a bridge-owned conversation/session object that encapsulates target creation, normalizer creation, system prompt setup, user-turn sending, empty-response handling, and user-turn count lookup. `LLMDriver` should depend on RAMPART types plus that bridge abstraction, not PyRIT classes. If the team intentionally wants PyRIT types to be public now, then update the architecture rules and boundary tests deliberately; leaving both stories in place will keep misleading future contributors.

🟡 Should Fix: stale `_pyrit` guidance remains after the package rename

> The PR says `_pyrit` moved to `pyrit_bridge`, but the repo instructions and payload generator docstring still point contributors at `_pyrit`. This matters because those instruction files are the source of truth for future Copilot and human reviews; right now they say one thing while the code does another.
> 
> **Recommendation:** update the guidance to pyrit_bridge and adjust the wording to match the intended new boundary. If PyRIT is now an openly acknowledged dependency, say that while still preserving the architectural boundary.

🟢 Small Note: direct `jinja2` usage relies on PyRIT’s dependency

rampart/drivers/llm.py
```diff
@@ -0,0 +1,353 @@
...
+
+    Termination is handled externally: the evaluator's early-stop
+    (on detection) or the execution loop's max_turns budget. The
+    driver never self-terminates — empty LLM responses raise
+    ``DriverError`` rather than returning None.
+
+    One driver instance = one driver-side conversation. Construct a
+    new driver per test. Use ``from_target`` for custom targets.
+
+    Args:
```
nina-msft: nit: args docstring missing "target" info

also is the args docstring needed for the class if there's already a docstring within `__init__`?

pyproject.toml
```diff
@@ -66,7 +69,7 @@ skip_empty = true
 
 [tool.pyright]
 pythonVersion = "3.11"
-typeCheckingMode = "strict"
+typeCheckingMode = "standard"
```
spencrr: this seems to still pass locally on strict - wdyt?

pyproject.toml
```diff
@@ -115,5 +118,8 @@ known-first-party = ["rampart"]
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.ruff.lint.pylint]
```
spencrr, line 126: ✅

rampart/drivers/llm.py
bashirpartovi: Yeah, I forgot to update that. Good call.

When we first created RAMPART, I was thinking we might use PyRIT for parts of the work, and if PyRIT lagged behind, we could add our own converters or similar pieces directly in RAMPART. But I don't think that's the right model anymore. If we need that capability, it's better to contribute it back to PyRIT and then consume it from RAMPART.

I'm trying to be more explicit that RAMPART uses PyRIT internally. That said, I still think it's valuable to keep a bridge layer so we don't expose PyRIT types directly through the RAMPART public API.

rampart/drivers/llm.py
bashirpartovi: We are importing and using PyRIT directly in the bridge. I think it makes sense to keep the bridge and contain most of the common PyRIT interactions there.

For example, converting `LLMConfig` to `PromptTarget` is something we need in multiple places, so it would be cleaner to keep that logic centralized in the bridge. If we think the bridge will become too much to maintain as the functionality grows, I'm also okay with removing it. What do you think?

But yeah, the coding standards need to be updated either way. I forgot to do that.

rampart/drivers/llm.py
bashirpartovi: yes good call, will add another validation here

rampart/drivers/llm.py
```diff
@@ -0,0 +1,353 @@
...
+            if self._llm is None:
+                msg = (
+                    "LLMDriver: no LLM config and no target — use "
+                    "from_target() or provide an LLMConfig."
+                )
+                raise DriverError(msg)
+            self._target = create_prompt_target(self._llm)
+            self._normalizer = PromptNormalizer()
+            self._target.set_system_prompt(
+                system_prompt=self._build_system_prompt(),
```
nina-msft: nit: Can we pull this out of the if/else block since its the same for both conditions?

rampart/core/execution.py
bashirpartovi: yes, great idea, I'll do that

rampart/drivers/llm.py
nina-msft: We could also potentially pull out the `if self._normalizer is None: self._normalizer = PromptNormalizer()` check too and then we would only need a check for `if self._target is None` to set the `_target` value and check for `_llm` existance.

rampart/drivers/llm.py
```diff
@@ -0,0 +1,353 @@
...
+
+        Raises:
+            DriverError: If the LLM call fails or returns an empty
+                response.
+
+        Args:
+            history: All agent-side turns so far (empty on first call).
+
+        Returns:
+            The next decision. Never returns None — termination is
```
nina-msft: nit: return type of `| None` is misleading based on docstring

rampart/drivers/llm.py
bashirpartovi: It doesn't really leak into the public API because consumers don't need to pass any PyRIT-specific types into RAMPART, except for the utility I added for `LLMDriver.from_target`.

That one is also just a convenience utility, but I do need to update the coding standards, which is probably why the reviewer flagged it (plus the test). Also, as you noticed, I moved the `_pyrit` boundary to `pyrit_bridge` to be more explicit about the fact that RAMPART uses PyRIT internally.

rampart/drivers/llm.py
nina-msft: haha twins

rampart/pyrit_bridge/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""PyRIT integration bridge."""
+
+from rampart.pyrit_bridge.llm_bridge import create_prompt_target, send_user_turn_async
+
+__all__ = [
+    "create_prompt_target",
```
nina-msft, line 13: potentially missing an export for `send_generation_request_async` which is used in `_generator.py`:L21? (https://github.com/microsoft/RAMPART/pull/33/changes#diff-ec270c79cdb960e4a6b1520c7c82d3aef761b1559e822a0914edf21bb1c52ba4R21)

rampart/drivers/llm.py
bashirpartovi: yes and yes, great suggestion, made the code way more readable, thanks Nina

.github/workflows/coverage.yml
bashirpartovi: The `--fail-under=0` is intentional, without it this summary step inherits the threshold from pyproject.toml and fails before the **Check coverage threshold step** even runs. That means when coverage drops, the job stops here and you never get the markdown report written to the summary (which is the whole point of this step).

With the change, there's a clean separation, this step always writes the report so we can see the numbers, and the next step is the single place that enforces the threshold. The coverage workflow still runs independently from lint+test, this doesn't change that.

rampart/drivers/llm.py
bashirpartovi: fixed the coding standards, this should be gone

rampart/core/execution.py
```diff
@@ -239,15 +242,31 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
...
-            logger.warning(
-                "%s during %s execution: %s",
-                error_type,
-                self.strategy_name,
-                exc,
-                exc_info=True,
-            )
+            is_expected = isinstance(exc, (InfrastructureError, DriverError))
+
+            if is_expected:
```
nina-msft: I would rephrase this list as custom domain exceptions vs standard python exceptions instead of expected / unexpected. We also raise `ValueError` `TypeError` `FileNotFoundError` and `pytest.UsageError` in the codebase.

Why wouldn't we handle all exceptions in the same way (e.g. why don't all of them fire ON_ERROR)?

bashirpartovi: You have a point. I originally was thinking that domain-specific exceptions are expected from execution strategies (meaning they would wrap runtime errors using Infrastructure Error) but we should not assume that. Will update it.

nina-msft (approved):
Great work Bashir!

nina-msft (approved):

PR #35, opened 2026-04-28
[FEAT]: Added userguide and API documentation

bashirpartovi:
# Add comprehensive documentation with MkDocs + mkdocstrings

## Summary

Adds a full documentation site for RAMPART using **MkDocs Material** and **mkdocstrings**. The docs are designed for engineers consuming RAMPART to write safety tests against their own agents — not for RAMPART maintainers.

### Documentation site infrastructure
- **`mkdocs.yml`** — MkDocs configuration with Material theme, mkdocstrings Python handler, navigation structure, and Mermaid diagram support
- **`docs/stylesheets/custom.css`** — Custom theme colors matching the RAMPART logo (dark charcoal + orange-amber)
- **`pyproject.toml`** — Added `docs` dependency group (`mkdocs`, `mkdocs-material`, `mkdocstrings[python]`) and setuptools package-find exclusion for `site/`

### Narrative documentation (hand-written)

| Section | Pages | Purpose |
|---------|-------|---------|
| **Getting Started** | Installation, Quickstart, pytest Integration | Setup repo, write first test, understand markers/fixtures |
| **Concepts** | Overview, Attacks, Probes, PyRIT Integration | Mental model, component diagram, lifecycle, evaluator polarity |
| **Attacks** | Index, XPIA | XPIA phases, inline vs surface-based injection, trigger options |
| **Probes** | Index, Behavioral | Behavioral probe usage, single/multi prompt, custom drivers |
| **Guides** | Authoring Tests, Configuration, Results & Reporting, CI Integration | Adapter/session/evaluator patterns, LLMConfig/Persona/AppManifest, report sinks, trial markers in CI |
| **Glossary** | 25 terms | Cross-referenced to API docs via mkdocstrings autorefs |

### API reference (auto-generated from docstrings)

12 pages using `:::` mkdocstrings directives — content is pulled directly from source docstrings at build time:

- Core Types, Core Protocols, Attacks, Probes, Evaluators, Drivers, Payloads, Surfaces, Converters, Reporting, pytest Plugin

### Source docstring fixes

Updated docstrings in 5 source files to improve rendered API docs:

- Converted RST `::` literal blocks to fenced ` ```python ` blocks for proper syntax highlighting in mkdocstrings
- Fixed `PromptDriver` docstring — removed inaccurate claim that all drivers are stateless and safe to reuse
- Files changed: `converter.py`, `payloads/__init__.py`, `payloads/template.py`, `reporting/json_file.py`, `pytest_plugin/plugin.py`

[...]

PR #36, opened 2026-04-30
[MAINT]: removing asyncio

richlundeen:
Remove all `@pytest.mark.asyncio decorators` from tests and drop the related instruction from unit test standards. The project uses `asyncio_mode = "auto"` (configured in `pyproject.toml`), which collects async tests automatically — making the decorators redundant.

### DIFF
--- a/.github/instructions/unit-tests-standards.instructions.md
+++ b/.github/instructions/unit-tests-standards.instructions.md
@@ -37,7 +37,6 @@ class TestParseConfig:
 ```
 
 ### Async Tests
-- Do NOT add `@pytest.mark.asyncio` — the project uses `asyncio_mode = "auto"` in `pyproject.toml`, so all `async def` tests are collected automatically
 - Async test method names MUST end with `_async`
 - Use `AsyncMock` instead of `MagicMock` when mocking async methods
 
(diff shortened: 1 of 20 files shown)


.github/instructions/unit-tests-standards.instructions.md
```diff
@@ -37,13 +37,12 @@ class TestParseConfig:
 ```
 
 ### Async Tests
-- Use `@pytest.mark.asyncio` decorator for all async test methods
+- Do NOT add `@pytest.mark.asyncio` — the project uses `asyncio_mode = "auto"` in `pyproject.toml`, so all `async def` tests are collected automatically
```
spencrr: I think we can just drop this - if it's not in any of our tests and there is a comment in pyproject.toml, that should be enough context. Can also enforce not using this via precommit or something. WDYT?

spencrr (approved):
Based on your comment and mine - easy change to remove some boilerplate. LGTM! https://github.com/microsoft/RAMPART/pull/35#discussion_r3157457930

bashirpartovi:
Thanks @rlundeen2 for this cleanup, makes total sense, just one request so we keep these PRs up to the standard, do you mind adding a short description to the PR ?

bashirpartovi (approved):

PR #37, opened 2026-05-07
[FEAT]: add architecture and test-execution-flow diagrams to the Concepts overview

bashirpartovi:
## Summary
Adds two new visual diagrams to the RAMPART documentation and rewrites the Concepts -> Overview page so that the diagrams flow naturally with the surrounding narrative. Also clears out a couple of pre‑existing doc‑warnings unrelated to the diagrams.

## What's changed
New diagrams (added to docs/images/):

1. `rampart-architecture-dark.svg` (+ .png fallback) - layered architecture showing PyRIT (L1) → RAMPART (L2) → Consumer Package (L3) → Agent Under Test (L4), and where the consumer's AgentAdapter and pytest suite plug in.
2. `rampart-execution-flow-dark.svg` (+ .png fallback) - request/response cycle for a single test run: Test → Attacks/Probes → AgentAdapter → Agent System and back.
3. Light variants are also included for completeness, but the docs currently embed the dark versions only.

`docs/concepts/overview.md` — restructured:

- Architecture at a Glance (new section) introduces the layered diagram and frames how the layers fit together. The pre‑existing component table now reads as a "zoom in" on the runtime path.
- Execution Lifecycle now leads with the execution‑flow diagram as the high‑level visual, then transitions into the existing mermaid sequence diagram as the per‑turn detail.
- The previous mermaid Component Model diagram (which the architecture image fully supersedes) was removed to avoid redundancy.

`docs/stylesheets/custom.css` - small CSS rule to make embedded SVG/PNG diagrams scale to the content column instead of honoring their intrinsic width/height attributes (which previously caused the diagrams to render either squeezed or overflowed).

Cleanup of pre‑existing doc warnings (unrelated to the diagrams, but they were noisy in mkdocs serve while iterating):
- Removed the orphaned docs/guides/ folder (authoring-tests.md, results-and-reporting.md). These files were byte‑identical duplicates of the canonical docs/usage/*.md versions already in the nav, were not linked from anywhere, and contained one broken cross‑link.

## How to verify
From the RAMPART directory:
```bash
uv sync --group docs
uv run mkdocs serve
```

Then open Concepts → Overview and check that:

- The "Architecture at a Glance" diagram renders at column width with readable text.
- The "Execution Lifecycle" diagram renders correctly above the existing mermaid sequence diagram.
- `mkdocs serve` no longer reports the orphaned‑pages warnings or the broken getting-started/pytest-integration.md#rampart_sinks link warning.

## Screenshots
image
image


richlundeen (approved):

docs/images/rampart-architecture-dark.png
spencrr, line 1: Looks great! I think we can drop the PNGs unless you want to reference them, but the diagrams are very helpful!

spencrr (approved):

docs/concepts/overview.md
```diff
@@ -27,26 +27,17 @@ RAMPART ships with the following probes; more will be added:
 
 ---
 
-## Component Model
+## Architecture at a Glance
 
-Every RAMPART test involves these components:
+RAMPART sits between your **consumer package** (the thin integration layer in your product team's repo) and the **agent under test**, building on PyRIT primitives underneath. The diagram below shows how the layers fit together and where your code plugs in:
```
nina-msft: nit: change "sits between" to "integrates within" since the diagram shows that the consumer package implements rampart fundamentals

docs/concepts/overview.md
```diff
@@ -27,26 +27,17 @@ RAMPART ships with the following probes; more will be added:
...
-        F[Attacks / Probes] --> E[Execution]
-        D[PromptDriver] --> E
-        EV[Evaluator] --> E
-        E --> R[Result]
-        R --> RP[ReportSink]
-    end
-    E -- "send_async()" --> S
-    E -- "inject()" --> SF
-```
+![RAMPART layered architecture: PyRIT (L1), RAMPART (L2), Consumer Package (L3), Agent Under Test (L4)](../images/rampart-architecture-dark.svg){ loading=lazy }
```
nina-msft, line 34: fyi i can see `{ loading=lazy }` as text, should we remove or define this elsewhere?

docs/concepts/overview.md
```diff
@@ -61,7 +52,13 @@ flowchart LR
 
 ## Execution Lifecycle
 
-Every execution follows a common lifecycle owned by [`BaseExecution`][rampart.core.execution.BaseExecution]:
+A single test run flows from your pytest test, through a RAMPART attack or probe, into your `AgentAdapter`, and out to the agent system — then back again as a `Result`:
+
+![RAMPART test execution flow: Test → Attacks/Probes → AgentAdapter → Agent System, with Pass/Fail returned](../images/rampart-execution-flow-dark.svg){ loading=lazy }
```
nina-msft, line 57: same flag for this `{ loading = lazy }` as above

nina-msft (approved):
small wording nit & tala look at my comments for the lazy load config, otherwise lgtm!

docs/concepts/overview.md
nina-msft: i don't see this in your screenshots, so for clarity I was viewing from the code here: https://github.com/bashirpartovi/RAMPART/blob/465638531315255a17f82169e45f6b10ea3cdf98/docs/concepts/overview.md

nina-msft: ah jk, this is a mkdocs-ism so github doesn't render correctly in the preview np

docs/images/rampart-architecture-dark.png
bashirpartovi: I'll keep the PNGs as reference since we use them a lot in slides and docx, so would be good to have a single source of truth.

docs/concepts/overview.md
bashirpartovi: done

nina-msft (approved):

PR #40, opened 2026-05-14
[DOCS]: Add Contributor Docs

nina-msft:
Co-authored-by: Copilot <copilot@github.com>

## Description

Creates contribution documentation for the project. This includes dev setup, code style & linting info, testing standards, architecture specifics, how to extend RAMPART, PR process, and release process.

_Other changes:_
1. This PR also removes the orphaned `docs\guides` folder, as these exist under `docs\usage`.
2. Replaces `pip install rampart` with `pip install git+https://github.com/microsoft/RAMPART.git`

## Breaking changes

None

## Checklist

- [ ] Tests added or updated for changes  --> N/A, doc-only
- [x] Documentation updated
- [x] `uv run mkdocs serve --strict` compiles **_(note to reviewer: run this command and view locally @ http://127.0.0.1:8000/)_**

Landing page:
image


docs/contributing/release-process.md
```diff
@@ -0,0 +1,228 @@
...
+git push origin releases/vx.y.z
+git tag -a vx.y.z -m "vx.y.z release" --force
+git push --tags --force
+```
+
+Rebuild the package after any cherry-pick and re-test.
+
+## 8. Publish to PyPI
+
+Create a PyPI account if you don't have one and ask another maintainer to add you to the `msft-rampart` project. Before publishing, have an API token scoped to the project ready (create one in your PyPI project settings).
```
nina-msft: note to update to align naming as appropriate

docs/contributing/code-style.md
```diff
@@ -0,0 +1,168 @@
...
+logger.info(f"Saved {len(payloads)} payloads to '{name}'")
+```
+
+### Docstrings
+
+Use Google-style docstrings with `Args:`, `Returns:`, and `Raises:` sections. Do not include example usage in docstrings.
+
+### PyRIT Bridge
+
+PyRIT-related logic should be grouped under `rampart/pyrit_bridge/`. PyRIT's import chain is heavy, so use lazy imports inside functions when wrapping PyRIT converters:
```
bashirpartovi: nit:
I would also include that no PyRIT-specific types should be exposed in the public API, maybe something like this:


> Common PyRIT integration logic should be grouped under `rampart/pyrit_bridge/`.
 Two rules apply:
> - **Do not expose PyRIT-specific types in RAMPART's public APIs.** Translate to/from RAMPART types at the bridge boundary so consumers don't have to depend on PyRIT directly.
> - **Defer heavy PyRIT imports with lazy imports inside functions.** PyRIT's import chain is heavy (~14s), so importing it at module top-level slows down RAMPART's startup. Use a local import where the PyRIT type is actually needed:

docs/contributing/architecture.md
```diff
@@ -0,0 +1,91 @@
+# Architecture for Contributors
+
+This page supplements the [Concepts Overview](../concepts/overview.md) with contributor-focused guidance: package layout, extension points, and key design decisions to understand before making changes. Read the **Concepts Overview** first for the component model and execution lifecycle.
+
+---
+
+## Package Layout
+
+The `rampart/` source tree is organized by concern: foundational types live in `core/`, while each extension point gets its own subpackage (`attacks/`, `probes/`, `evaluators/`, `drivers/`, `converters/`, `surfaces/`, `payloads/`, `reporting/`). The `pyrit_bridge/` package isolates all PyRIT interaction (see [PyRIT Isolation](#pyrit-isolation) below), and `pytest_plugin/` provides the pytest integration.
```
bashirpartovi: We don't want to say pyrit_bridge isolates PyRIT related interactions, it does not. It only groups PyRIT-related logic in a common place. Maybe something like this:
```markdown
The `pyrit_bridge/` package groups common PyRIT integration code (see [PyRIT Bridge](#pyrit-bridge) below), and `pytest_plugin/` provides the pytest integration.
```

bashirpartovi (commented):
Looks great, just a few nits :)

bashirpartovi (approved):
LGTM, approved

docs/contributing/release-process.md
spencrr: nit also here

docs/contributing/extending-rampart.md
```diff
@@ -0,0 +1,357 @@
...
+    AgentAdapter,
+    BaseExecution,
+    Evaluator,
+    ExecutionEventHandler,
+    PromptDriver,
+    Result,
+    Turn,
+    resolve_as_attack,
+)
+from rampart.core.execution import evaluate_turn_async
```
spencrr: Ai find:

But evaluate_turn_async is re-exported from rampart.core (rampart/core/__init__.py), and docs/contributing/code-style.md explicitly says to "Import from the package root … not from internal file paths." Move evaluate_turn_async into the first import block. (Or, if there's a deliberate reason — e.g., the production _xpia.py does the same — call that out; but the production file isn't bound by the same rule because the rule applies to "the package root when the symbol is exported there".)

docs/contributing/release-process.md
```diff
@@ -0,0 +1,239 @@
...
+| Component | Increment when |
+|-----------|---------------|
+| **MAJOR** | Breaking changes to the public API |
+| **MINOR** | New features, new attack/probe types, backward-compatible additions |
+| **PATCH** | Bug fixes, documentation corrections, dependency updates |
+
+!!! note "Pre-1.0 stability"
+    While RAMPART is below `1.0`, minor version bumps may include breaking changes. The API is stabilizing but not yet frozen. The first stable release will be `1.0.0`.
+
+In line with PyPA [versioning guidance](https://packaging.python.org/en/latest/discussions/versioning/), `main` may carry a `.dev0` suffix between releases (e.g., `0.2.0.dev0`). Release commits drop the suffix; the post-release bump on `main` reintroduces it for the next version.
```
spencrr: note that we recommend SHA for anyone following main that still wants to pin? unsure about best guidance to users here.

docs/contributing/release-process.md
```diff
@@ -0,0 +1,239 @@
...
+
+Commit your changes to a release branch and push the tag:
+
+```bash
+git checkout -b releases/vx.y.z
+git commit -am "release vx.y.z"
+git push origin releases/vx.y.z
+git tag -a vx.y.z -m "vx.y.z release"
+git push --tags
+```
```
spencrr, line 67: let's change this to not have patch branches, only minor version branches and then tag patches for release. wdyt?

docs/contributing/pull-requests.md
```diff
@@ -0,0 +1,103 @@
...
+    - Documentation is updated for user-facing changes
+    - Code follows the project's [coding standards](code-style.md)
+    - The change is well-scoped and doesn't introduce unnecessary complexity
+
+!!! tip
+    Open an [issue](https://github.com/microsoft/RAMPART/issues) before starting work on large features or architectural changes. This helps align on approach before investing time in implementation.
+
+## Stale Pull Requests
+
+If a pull request has no activity for an extended period, maintainers may check in with the author. If there is no response within 14 days, maintainers may reassign the work to ensure progress continues.
```
spencrr, line 103: are we committing to 14 days or kind of just a soft expectation?

docs/contributing/pull-requests.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Lint & Type Check
+
+- **Ruff check** — all lint rules pass
+- **Ruff format** — code is properly formatted
+- **Pyright** — strict type checking passes
+
+### Tests
+
+- Unit tests pass on **Python 3.11, 3.12, and 3.13**
```
spencrr: link to pyproject toml and maybe ci to keep in sync if we ever upgrade?

docs/contributing/release-process.md
spencrr: do we want to track this or just follow up immediately following pypi transfer?

spencrr (approved):

docs/contributing/pull-requests.md
nina-msft: seemed reasonable to me that we would get to it within this time frame, but we can adjust later if this becomes untenable :)

docs/contributing/release-process.md
spencrr: Not necessarily this but perhaps:

```suggestion
In line with PyPA [versioning guidance](https://packaging.python.org/en/latest/discussions/versioning/), `main` may carry a `.dev0` suffix between releases (e.g., `0.2.0.dev0`). Release commits drop the suffix; the post-release bump on `main` reintroduces it for the next version.

!!! note Version pinning
Consider pinning to a specific revision or version tag over the active main branch or `.devN`  versions.
```

docs/contributing/extending-rampart.md
nina-msft: d'oh! thanks :D didn't check on the example code lol

docs/contributing/release-process.md
nina-msft: removed this section since it might be confusing and is already covered in step 9

docs/contributing/release-process.md
nina-msft: won't change for this pr - discussed in dm

spencrr (approved):

PR #54, opened 2026-05-20
fix: fix runtime errors and improve partial-failure handling

xyplex3:
Key Changes:

  - Fixed default_factory fields using non-callable generic aliases (e.g., list[Turn], dict[str, Any]) that raise TypeError at runtime — replaced with bare list and dict
  - Fixed _validate_collection_name calls misplaced before docstrings in PayloadStore, making them syntactically dead code — moved to correct position after docstrings and
  extended to all mutating methods
  - Improved PayloadGenerator to survive partial LLM failures — uses return_exceptions=True and raises DriverError only when all variants fail

  Changed:

  - default_factory callables corrected across Result, Payload, ToolCall, SideEffect, Response, Request, EvalResult, and EvalContext — generic aliases like list[Turn] are not
  callable and would raise TypeError on instantiation (rampart/core/result.py, rampart/core/types.py)
  - _validate_collection_name placement fixed in PayloadStore.save and PayloadStore.load, where calls appeared before the docstring and were never reached; validation also added
  to exists, delete, and get_manifest for consistency (rampart/payloads/_store.py)
  - asyncio.gather in PayloadGenerator._generate_variants now passes return_exceptions=True, filters successful results, warns on partial failure, and raises DriverError only if
  every call fails — previously a single failure would propagate uncaught (rampart/payloads/_generator.py)
  - _build_summary in _xpia.py and PayloadFormat.extension in types.py refactored from if-chains to match/case with explicit wildcard branches for exhaustiveness
  - shutil.move call updated to pass Path objects directly instead of wrapping in str() — unnecessary since Python 3.9 (rampart/payloads/_store.py)
  - Docstrings for Payload.__post_init__, Request.__post_init__, and EvalContext.current_turn expanded to document raised exceptions; strategy_name docstring updated to imperative
   form

### DIFF
--- a/rampart/attacks/_xpia.py
+++ b/rampart/attacks/_xpia.py
@@ -89,7 +89,7 @@ class XPIAExecution(BaseExecution):
 
     @property
     def strategy_name(self) -> str:
-        """Identifies this as an XPIA execution in results and reports."""
+        """Return the strategy name used in results and reports."""
         return "xpia"
 
     async def _execute_async(self, *, adapter: AgentAdapter) -> Result:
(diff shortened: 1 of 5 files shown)


xyplex3:

PR #56, opened 2026-05-20
[MAINT]: Add Python 3.14 Support

spencrr:
## Description

Adds Python 3.14 to RAMPART's officially supported/tested versions.

### Changes
- `pyproject.toml`: added `Programming Language :: Python :: 3.14` classifier.
  `requires-python = ">=3.11"` already permitted 3.14, no constraint change needed.
- `.github/workflows/ci.yml`: added `3.14` to the test matrix.
- `.github/ISSUE_TEMPLATE/bug_report.yml`: added `3.14` to the Python version dropdown.
- `uv.lock`: refreshed (no resolution changes).

### Verification
- `uv lock` resolves all 163 transitive dependencies cleanly on 3.14.
- `uv sync --python 3.14` installs cleanly
- Full test suite passes locally on CPython 3.14.3

### Notes on what we deliberately did *not* change
- No upper cap on `requires-python` — per [PyPA guidance](https://iscinumpy.dev/post/bound-version-constraints/), capping propagates pain to downstream consumers.
- Pyright `pythonVersion` stays pinned to the floor (`3.11`) so type-checks catch code that wouldn't run on the oldest supported version.
- Ruff has no explicit `target-version`; it derives from `requires-python`, which is correct.

### Related
PyRIT added 3.14 support in microsoft/PyRIT#1130, which is what unblocked picking it up here (pyrit is our heaviest transitive constraint).

## Breaking changes

None

## Checklist

- [X] Tests added or updated for changes
- [X] Documentation updated


uv.lock
```diff
@@ -2965,7 +2970,7 @@ wheels = [
 
 [[package]]
 name = "rampart"
-version = "0.1.0"
+version = "0.1.1.dev0"
```
nina-msft, line 2973: Thanks for adding this!

nina-msft (approved):

uv.lock
spencrr: whoopsie on the release! I know #53 added this too 😉

PR #58, opened 2026-05-21
[DOCS]: expand README with installation, quick start, and documentation links

maiconburn:
## Description

The current README ships a single descriptive paragraph followed immediately by the Trademarks block, which leaves no on-ramp for someone landing on the GitHub repo. Since v0.1.0 is already published on PyPI, the README ought to surface the install command, a runnable Quick Start, and a navigable map of the docs site.

This PR adds, in order, between the existing description and the Trademarks block:

* **Installation** with the standard `pip install rampart` command.
* **Quick Start** using the canonical XPIA snippet lifted verbatim from `docs/index.md`, so the README example cannot drift from the docs site.
* **Documentation** with a short bullet list pointing at the five key entry points on microsoft.github.io/RAMPART (Getting Started, Concepts, XPIA Attack, Behavioral Probe, API Reference).
* **Contributing** with a one-line pointer to the Contributing Guide plus the Code of Conduct and CLA links.

The Trademarks block moves to the bottom (conventional position).

No code, no tests, no docs site content changed. Each new link target was checked against the on-disk docs tree (`docs/getting-started/index.md`, `docs/concepts/overview.md`, `docs/attacks/xpia.md`, `docs/probes/behavioral.md`, `docs/api/index.md`) and the `mkdocs.yml` nav, so all five links resolve at the published docs site.

## Breaking changes

None.

## Checklist

- [x] \`pre-commit run --all-files\` passes (no Python files changed; ruff and pyright are no-ops on markdown).
- [x] Tests added or updated for changes (n/a, README-only).
- [x] Documentation updated (this is the documentation update).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,40 @@
 
 RAMPART provides a structured, developer-friendly way to write and run safety and security tests for AI agents -- covering **adversarial attacks**, **benign failures**, and a broad range of **harm categories**, all with evaluation-driven assertions and seamless integration with [pytest](https://docs.pytest.org/).
 
+## Installation
+
+```bash
+pip install rampart
+```
+
+## Quick Start
+
+```python
+result = await Attacks.xpia(
+    trigger="Summarize the Q3 reports",
+    evaluator=ToolCalled("send_email"),
+    inject=handle,
+).execute_async(adapter=my_agent)
+
+assert result, result.summary
+```
+
+See [Getting Started](https://microsoft.github.io/RAMPART/getting-started/) for the adapter pattern, installation prerequisites, and a complete first test.
+
+## Documentation
+
+Full documentation lives at **[microsoft.github.io/RAMPART](https://microsoft.github.io/RAMPART/)**:
+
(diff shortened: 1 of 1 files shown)


maiconburn:

nina-msft (changes requested):
Thanks for this contribution! I agree with Bashir's feedback below in this case. We'd like to reduce duplication where possible and the installation / quick start instructions are easy to find from the website's landing page.

We hope to see you contribute again in the future!

bashirpartovi:
@maiconburn thank you for helping to improve our documentation; however, I feel that this PR is probably not necessary for the following reasons:
1. The quick start guide you have included attempts to duplicate what we have in `docs/` and if we ever change once place, we need to change the other which could become confusing, I'd rather keep it in `docs/` the way we have it now. 
2. Documentation is already part of "about" which is a standard place to include general docs in Github. 
3. Contribution is already a tab at the top so no need to include that. 
@nina-msft @spencrr what's your take here? 
image

PR #59, opened 2026-05-21
[security] fix(payloads): validate store collection names

Hinotoi-agent:
## Summary

This PR hardens `PayloadStore` collection-name handling so all public collection lookup and removal APIs enforce the same trust boundary as `save()` and `load()`.

Before this change, `save()` and `load()` rejected traversal-like collection names, but `exists()`, `delete()`, and `manifest()` used the caller-provided collection name directly in filesystem paths. A malformed collection name such as `..` or `../outside` could therefore resolve outside the configured payload store root for these management APIs.

## Security issues covered

- Payload store collection-name path traversal in collection management APIs.

## Before this PR

- `save()` and `load()` validated collection names.
- `exists()`, `delete()`, and `manifest()` did not validate collection names before joining them with the store root.
- `delete("..")` could attempt to remove a directory outside the payload store root.
- `manifest("..")` could read a `manifest.json` outside the payload store root if present.

## After this PR

- `exists()`, `delete()`, and `manifest()` validate collection names before filesystem access.
- Collection names are restricted to filename-safe identifiers using letters, numbers, dots, underscores, or hyphens.
- Special dot entries (`.` and `..`), path separators, control characters, and other unsafe filesystem characters are rejected consistently.

## Why this matters

Payload collections are file-backed and may be managed from test helpers, fixtures, or higher-level tooling. Collection names should remain identifiers, not filesystem paths. Keeping every public collection API behind the same validation boundary prevents accidental or malicious traversal outside the configured payload cache root.

The highest-impact path was `delete()`: if a caller passed an unsafe collection name, the method could target a parent or sibling directory instead of a payload collection. Even when this raises on some platforms, reaching a destructive filesystem operation with an unvalidated traversal name is unsafe behavior.

## Attack flow

1. A caller obtains influence over a payload collection name used by store management code.
2. The caller supplies a traversal-like name such as `..`.
3. `PayloadStore.delete()` or `PayloadStore.manifest()` joins that value with the configured store root without validation.
4. The resulting path resolves outside the intended payload store boundary.
5. The operation can attempt deletion or manifest reads outside the payload collection directory.

## Affected code

- `rampart/payloads/_store.py`
  - `PayloadStore.exists()`
  - `PayloadStore.delete()`
  - `PayloadStore.manifest()`
  - `_validate_collection_name()`

## Root cause

- Collection-name validation was applied only to some public methods.
- Management methods reused raw caller-provided names directly in filesystem paths.
- The previous validator rejected path separators and dot entries but did not define a single positive allowlist for safe collection identifiers.

## CVSS assessment

- Issue: Payload store collection-name path traversal in management APIs
- CVSS v3.1: 5.5 Medium
- Vector: `CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:L/I:L/A:L`
- Rationale: exploitation requires local/API-level ability to provide a collection name in the test process, but the affected operations can read or attempt to delete filesystem paths outside the configured payload store root.

## Safe reproduction steps

On vulnerable code, a focused regression test demonstrates the unsafe boundary without touching real user files:

```python
root = tmp_path / "store"
root.mkdir()
sentinel = tmp_path / "sentinel.txt"
sentinel.write_text("do not delete")
store = PayloadStore(root=root)

store.delete("..")
```

The new test expects this to raise `ValueError` before any destructive filesystem operation is reached.

A second regression covers metadata reads:

```python
root = tmp_path / "store"
root.mkdir()
(tmp_path / "manifest.json").write_text('{"collection": "outside"}')
store = PayloadStore(root=root)

store.manifest("..")
```

The fixed behavior rejects the unsafe collection name instead of resolving a manifest outside the store root.

## Expected vulnerable behavior

- `exists("../outside")` does not reject the unsafe name.
- `delete("..")` reaches `shutil.rmtree()` on a path outside the intended collection namespace.
- `manifest("..")` can read `../manifest.json` relative to the store root if that file exists.

## Changes in this PR

- Adds a positive collection-name allowlist: `[A-Za-z0-9._-]{1,128}`.
- Rejects `.` and `..` explicitly.
- Applies collection-name validation to `exists()`, `delete()`, and `manifest()`.
- Adds regression tests for traversal rejection across read, existence, and delete management paths.

## Files changed

- `rampart/payloads/_store.py`
  - Adds the collection-name allowlist and validation calls in management APIs.
- `tests/unit/payloads/test_store.py`
  - Adds regression coverage for traversal attempts against `exists()`, `delete()`, and `manifest()`.

## Maintainer impact

This is intentionally narrow. Existing collection names made from letters, numbers, dots, underscores, and hyphens continue to work. Names containing path separators or other filesystem-special characters now fail early with `ValueError`.

The patch does not change payload serialization format, artifact layout, or normal save/load behavior for valid collection names.

## Fix rationale

A positive identifier allowlist is safer than trying to enumerate every platform-specific path edge case. It also keeps payload collection names portable across operating systems and consistent across all public `PayloadStore` APIs.

Applying validation at method entry preserves the intended abstraction: callers pass a collection identifier, and the store decides how that identifier maps to paths under its root.

## Type of change

- [x] Security hardening
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] Breaking change

## Test plan

Focused regression tests:

```bash
uv run pytest tests/unit/payloads/test_store.py -q
```

Result:

```text
16 passed
```

Full test suite:

```bash
uv run pytest -q
```

Result:

```text
451 passed
```

Additional checks:

```bash
uv run ruff check rampart/payloads/_store.py tests/unit/payloads/test_store.py
uv run ruff format --check rampart/payloads/_store.py tests/unit/payloads/test_store.py
uv run pyright rampart/payloads/_store.py tests/unit/payloads/test_store.py
python -m compileall -q rampart/payloads/_store.py tests/unit/payloads/test_store.py
git diff --check
```

Result: all passed.

## Disclosure notes

This PR is limited to collection-name validation in `PayloadStore` management APIs. It does not claim to address unrelated payload artifact path handling or external storage integrations.


bashirpartovi:
Hi @Hinotoi-agent 

Do you mind fixing the conflict?

bashirpartovi (changes requested):
Generated by Copilot

I found two blocking issues on head `fa96b8d`.

### 1. Windows filename aliases can delete another collection

**Location:** `rampart/payloads/_store.py:38`, `rampart/payloads/_store.py:189-195`, `rampart/payloads/_store.py:218-229`

The allowlist accepts names ending in a period. On Windows, `victim.` resolves to the same directory as `victim`. Against the current head, `exists("victim.")` returned `True`, and `delete("victim.")` removed the `victim` collection. This means one accepted identifier can target a different collection, so the intended filesystem boundary is incomplete.

The validator also accepts reserved Windows device basenames such as `CON`, `NUL`, `COM1`, and extension forms such as `CON.txt`, despite describing the names as portable and filename-safe.

**Recommendation:** Reject trailing periods and spaces, and reject reserved Windows device basenames case-insensitively, including extension forms. Please add regression tests for `victim.`, repeated trailing periods, `CON`, `NUL`, and `CON.txt`.

### 2. The stricter grammar strands existing public collections

**Location:** `rampart/payloads/_store.py:173-186`, `rampart/payloads/_store.py:218-229`, `rampart/payloads/__init__.py:49-53`

`PayloadStore` is publicly exported. The previous validator accepted simple collection names containing spaces, Unicode, or more than 128 characters. The new ASCII/128-character allowlist rejects those names across lookup and removal APIs, while `list_collections()` still returns existing directories with those names.

For example, an existing `team payloads` collection is returned by `list_collections()`, but passing that returned identifier to `exists()`, `load()`, `manifest()`, or `delete()` raises `ValueError`. Such collections become inaccessible through the API even though this PR is marked as non-breaking.

**Recommendation:** Prefer preserving the prior accepted character set while rejecting actual traversal and platform aliases. If the strict ASCII/128-character contract is intentional, document this as a breaking change and provide a collision-aware migration path; `list_collections()` must not return identifiers that the other public APIs reject.

The original literal `.` / `..` and separator traversal is fixed, and the existing test suite passes. These two cases are missing security and compatibility coverage, so I am requesting changes until they are addressed.

Hinotoi-agent:
Addressed both requested changes in `62eaf0a`:

- preserved the prior collection-name character set (including spaces, Unicode, and names longer than 128 characters) while continuing to reject empty names, separators, and `.` / `..`;
- rejected Windows trailing-dot/space aliases and reserved device basenames case-insensitively, including extension forms such as `CON.txt`;
- added regression coverage showing `victim.`, repeated trailing periods, and a trailing space cannot delete `victim`, reserved names are rejected, and compatible legacy names round-trip through save/list/exists/load/manifest/delete.

Validation:
- `uv run pytest tests/unit/payloads/test_store.py -q` — 27 passed
- full `uv run pytest -q` — 644 passed, 5 skipped
- Ruff check/format, ty, compileall, and `git diff --check` passed
- Docker focused run on Python 3.12 — 27 passed

rampart/payloads/_store.py
```diff
@@ -211,14 +222,24 @@ def manifest(self, name: str) -> dict[str, Any]:
...
-                or is a relative-path component (``.`` / ``..``).
+            ValueError: If the name is not a portable directory name.
         """
-        if not name or "/" in name or "\\" in name or name in {".", ".."}:
-            msg = f"Invalid collection name: {name!r}. Must be a simple directory name."
+        basename = name.partition(".")[0].upper()
+        is_unsafe_path = (
+            not name
+            or name in {".", ".."}
+            or any(separator in name for separator in ("/", "\\"))
```
nina-msft, line 234: This still allows Windows drive-relative names such as `C:outside`. With the default relative root on Windows, `self._root / "C:outside"` resolves to `<cwd>\outside`, so `manifest()` can read and `delete()` can remove a collection outside `.rampart\payloads`.

One way to close this gap is to use Windows path semantics explicitly:

```python
from pathlib import Path, PureWindowsPath

is_unsafe_path = (
    not name
    or name in {".", ".."}
    or any(separator in name for separator in ("/", "\\"))
    or bool(PureWindowsPath(name).drive)
)
```

Please also add regression coverage for `C:` and `C:outside`, including a destructive API such as `delete()`.

nina-msft (changes requested):

PR #70, opened 2026-06-02
[DOCS]: Add portable regression receipt example to results-and-reporting

### LINKED ISSUE #63: Docs: example pattern for agent safety regression receipts in reporting sinks
## What I think is confusing or missing

RAMPART already has the pieces for high-quality regression evidence:

- `Result` captures `status`, `summary`, `turns`, `harm_category`, `strategy`, `injections`, and `metadata`
- `TestRunReport` carries run-level `metadata`
- `JsonFileReportSink` serializes per-turn prompts, tool calls, side effects, eval rationale, and response metadata

I may be missing an intended pattern, but I did not find docs showing how a team should turn those primitives into a stable **agent safety regression receipt** that can live in CI and be compared across runs.

Right now `docs/usage/results-and-reporting.md` explains what the framework emits. I think a small example could make it clearer how to preserve one repeatable artifact for scenarios like:

- "this was a prompt-injection regression"
- "this was a benign-failure probe"
- "this run should prove the mitigation stayed in place after a fix"

My concern is that teams may otherwise invent incompatible reporting schemas even though the framework already has the underlying fields.

A concrete docs example seems especially useful for people using RAMPART as a CI gate rather than only as an exploratory red-team tool.

## Possible docs direction

I think a small addition to `docs/usage/results-and-reporting.md` could be enough:

- a short section such as **Portable regression receipt** or **Recommended report metadata**
- one example that uses existing `report.metadata` + `result.metadata` fields, without introducing a new core type first
- a note tying the example to the built-in JSON sink so people can see what should be persisted

Something along these lines:

```json
{
  "scenario_id": "xpia-sharepoint-basic-001",
  "threat_class": "prompt_injection",
  "benign_or_adversarial": "adversarial",
  "agent_adapter": "MyCopilotAdapter",
  "fixture_ref": "tests/safety/test_sharepoint_xpia.py::test_hidden_instruction_blocked",
  "expected_safe_behavior": "ignore hidden instruction and avoid exfiltration",
  "evaluator_version": "llm_judge:v1",
  "verdict": "safe",
  "trace_ref": "ci://build/1842/artifacts/run_report_2026-05-26T14-30-00.json",
  "ci_run_url": "https://github.com/org/repo/actions/runs/123456789",
  "mitigation_ref": "docs/security/sharepoint-hardening.md#hidden-instruction-filter"
}
```

I think this can stay docs-only at first. The main value would be giving users a canonical pattern for what to put in `metadata`, what to preserve from `Result`, and how to keep mitigation regressions reviewable over time.

sumit1kr:
Closes #63

Adds a "Portable Regression Receipt" section to `docs/usage/results-and-reporting.md` 
showing how to use existing `report.metadata` and `result.metadata` fields to produce 
a stable, diffable CI artifact tied to `JsonFileReportSink`.

No new Python types introduced — docs only.

### DIFF
--- a/docs/usage/results-and-reporting.md
+++ b/docs/usage/results-and-reporting.md
@@ -70,6 +70,46 @@ sink = JsonFileReportSink(output_dir=Path(".report"))
 
 Output: `.report/run_report_2026-04-25T14-30-00.json`
 
+---
+
+## Portable Regression Receipt
+
+For CI gating, capture a stable JSON artifact that teams can diff across runs. Use
+`report.metadata` for run-level context (what was tested) and `result.metadata`
+for scenario-level facts (what should stay stable across time). Persist the
+`JsonFileReportSink` output as your regression receipt.
+
+```json
+{
+    "report": {
+        "metadata": {
+            "scenario_id": "xpia-login-001",
+            "threat_class": "credential_exfiltration",
+            "benign_or_adversarial": "adversarial",
+            "agent_adapter": "AcmeAgentAdapter:v2",
+            "fixture_ref": "tests/fixtures/login_prompt.yaml#v4",
+            "ci_run_url": "https://ci.example.com/runs/94821"
+        }
+    },
+    "results": [
+        {
+            "metadata": {
+                "expected_safe_behavior": "never reveal a password or token",
+                "evaluator_version": "response_contains@1.4.2",
+                "verdict": "UNSAFE",
+                "trace_ref": "memory://conv/9f8a6",
+                "mitigation_ref": "SEC-1234"
+            },
+            "status": "UNSAFE",
+            "safe": false,
+            "strategy": "xpia",
+            "harm_category": "DATA_EXFILTRATION",
+            "summary": "Agent leaked a token in response to a prompt injection."
+        }
+    ]
+}
+```
+
 ### Custom Sinks
 
 Implement the [`ReportSink`][rampart.reporting.sink.ReportSink] protocol:
--- a/docs/usage/results-and-reporting.md
+++ b/docs/usage/results-and-reporting.md
@@ -75,38 +75,55 @@ Output: `.report/run_report_2026-04-25T14-30-00.json`
 ## Portable Regression Receipt
 
 For CI gating, capture a stable JSON artifact that teams can diff across runs. Use
-`report.metadata` for run-level context (what was tested) and `result.metadata`
-for scenario-level facts (what should stay stable across time). Persist the
-`JsonFileReportSink` output as your regression receipt.
+`result.metadata` for scenario-level facts (what should stay stable across time)
+and run-level context (what was tested). Persist the `JsonFileReportSink` output
+as your regression receipt.
+
+!!! note
+    `report.metadata` (run-level metadata) is not currently included in the serialized output of `JsonFileReportSink`. If you need to track run-level context (such as `scenario_id` or `ci_run_url`) in the output, nest these fields under `result.metadata` inside the individual test results.
 
 ```json
 {
-    "report": {
+  "total_runs": 1,
+  "passed": 0,
+  "failed": 1,
+  "undetermined": 0,
+  "errors": 0,
+  "duration_seconds": 1.23,
+  "population_summary": {
+    "total_runs": 1,
+    "safe_count": 0,
+    "unsafe_count": 1,
+    "error_count": 0,
+    "attack_success_rate": 1.0,
+    "safety_pass_rate": 0.0
+  },
+  "by_harm_category": {
+    "DATA_EXFILTRATION": [
+      {
+        "safe": false,
+        "status": "UNSAFE",
+        "summary": "Agent leaked a token in response to a prompt injection.",
+        "harm_category": "DATA_EXFILTRATION",
+        "strategy": "xpia",
+        "duration_seconds": 1.23,
         "metadata": {
-            "scenario_id": "xpia-login-001",
-            "threat_class": "credential_exfiltration",
-            "benign_or_adversarial": "adversarial",
-            "agent_adapter": "AcmeAgentAdapter:v2",
-            "fixture_ref": "tests/fixtures/login_prompt.yaml#v4",
-            "ci_run_url": "https://ci.example.com/runs/94821"
-        }
-    },
-    "results": [
-        {
-            "metadata": {
-                "expected_safe_behavior": "never reveal a password or token",
-                "evaluator_version": "response_contains@1.4.2",
-                "verdict": "UNSAFE",
-                "trace_ref": "memory://conv/9f8a6",
-                "mitigation_ref": "SEC-1234"
-            },
-            "status": "UNSAFE",
-            "safe": false,
-            "strategy": "xpia",
-            "harm_category": "DATA_EXFILTRATION",
-            "summary": "Agent leaked a token in response to a prompt injection."
-        }
+          "scenario_id": "xpia-login-001",
+          "threat_class": "credential_exfiltration",
+          "benign_or_adversarial": "adversarial",
+          "agent_adapter": "AcmeAgentAdapter:v2",
+          "fixture_ref": "tests/fixtures/login_prompt.yaml#v4",
+          "ci_run_url": "https://ci.example.com/runs/94821",
+          "expected_safe_behavior": "never reveal a password or token",
+          "evaluator_version": "response_contains@1.4.2",
+          "verdict": "UNSAFE",
+          "trace_ref": "memory://conv/9f8a6",
+          "mitigation_ref": "SEC-1234"
+        },
+        "turns": []
+      }
     ]
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sumit1kr:

gitrishiom:
@sumit1kr   

Apologies for the slow response, and thanks for taking this on.

I agree with the docs-only direction here. I also agree with keeping the broader runtime-evidence / production-receipt idea from #63 out of scope for this PR; this section works best as test-time CI evidence using RAMPART’s existing reporting surfaces.

One small thing I noticed while re-checking the current sink behavior: `TestRunReport` has `metadata`, but `JsonFileReportSink._serialize_report()` does not currently include `report.metadata` in the emitted JSON. It serializes the aggregate fields, `population_summary`, `by_harm_category`, and each result’s `metadata`.

Could you update the docs example to mirror the current `JsonFileReportSink` output shape and put the stable receipt fields under `result.metadata` / the relevant `by_harm_category` result entry?

If maintainers would rather support run-level metadata in the built-in JSON output, then the alternate path would be a tiny serialization + test change for `report.metadata`, but my preference is to keep this PR docs-only and avoid scope creep.

Once that small alignment is made, I’m happy to take another pass on the diff. Thanks again for moving this forward.

sumit1kr:
@gitrishiom Updated — moved the receipt fields under `result.metadata` (nested inside the relevant `by_harm_category` entry) to match the current `JsonFileReportSink` output shape, and added a note clarifying that `report.metadata` isn't currently serialized. Pushed as a follow-up commit. Ready for another pass whenever you get a chance — thanks for the detailed review!

@spencrr tagging again since this is ready for maintainer review/approval whenever you have bandwidth.

gitrishiom:
Thanks, this addresses the concern.

The example now matches the current `JsonFileReportSink` output shape, and the note about `report.metadata` not being serialized is helpful. From my side, this looks aligned with the docs-only scope for #63.

I’ll leave final review/approval to @spencrr . Thanks again for the quick update.

nina-msft:
Thanks for picking this up @sumit1kr, and apologies for the delayed review! This is genuinely useful and it does hit the intent of #63 (giving teams a canonical way to use `result.metadata` for regression evidence). Two things I'd like to resolve before it lands:

**1. Placement.** The new `## Portable Regression Receipt` currently sits between `### JsonFileReportSink` and `### Custom Sinks`, so a top-level heading splits the *Report Sinks* section in two. Could we demote it to `####` under the sink subsection (or move it out below Report Sinks)? Since it's fundamentally about `result.metadata`, it reads more naturally as a subsection there.

**2. Reframe the example around `result.metadata` rather than reproducing sink output.** As written, the JSON mixes framework-emitted fields (`status`, `safe`, `strategy`) with user metadata and mirrors the `JsonFileReportSink` envelope — which risks going stale if we change the sink, and a few details don't match today's output (there's no `report`/`results` wrapper, `report.metadata` isn't serialized, `status` serializes lowercase, and results are actually nested under `by_harm_category`). I think it's clearer to show the fields being *attached* to the Result, so it's obvious they come from `result.metadata`:

 ```python
result = await Attacks.xpia(...).execute_async(adapter=my_adapter)

# Scenario-level facts you want stable across runs — pick the keys your team needs
 result.metadata.update({
     "scenario_id": "xpia-login-001",
     "threat_class": "credential_exfiltration",
     "expected_safe_behavior": "never reveal a password or token",
     "evaluator_version": "response_contains@1.4.2",
     "mitigation_ref": "SEC-1234",
 })

 assert result, result.summary
 ```

These keys travel with the `Result` into whatever sink is wired up. With the built-in `JsonFileReportSink`, they show up on that result's `metadata` object (results are grouped under `by_harm_category`). Worth noting for readers: the framework also contributes a couple of keys to `metadata` automatically — `test_name`, plus `harm_category` when the test carries a `@pytest.mark.harm` marker — so the persisted `metadata` will contain slightly more than the snippet sets.

That keeps the doc teaching the intended API without coupling it to the sink's serialization format. The field names are all illustrative anyway, since the receipt schema is inherently per-team. What do you think?

nina-msft (changes requested):
See block comment below for requested updated :-) Thanks for your contribution!

sumit1kr:
@nina-msft Thanks for the detailed review! Both changes addressed:

1. Moved `## Portable Regression Receipt` below the full Report Sinks section so it no longer splits the subsections.
2. Replaced the JSON envelope with a `result.metadata.update()` snippet, updated the prose to explain sink behaviour, and added the note about framework-contributed keys (`test_name`, `harm_category`).

Ready for another look whenever you get a chance.

nina-msft:
@sumit1kr - I made a few tweaks in my [latest commit](https://github.com/microsoft/RAMPART/pull/70/commits/a96da19b2722fbc61aa049d68effef68352dfb87) - let me know what you think!

* Moved the section — "Portable Regression Receipt" relocated from between the sink subsections to the end of the page, after ## TestRunReport. This fixes the heading-splitting issue and means by_harm_category is now introduced before it's referenced.

* Rewrote the intro — drops "Persist the JsonFileReportSink output as your regression receipt" in favor of routing both scenario-level and run-level context into result.metadata.

* Fixed the sink-accuracy sentence — now "any sink can persist them… A custom sink only records them if its emit_async reads result.metadata," instead of the previously over-general "travel into whatever sink is wired up." that I suggested

* Deleted the confusing second note — the dangling report.metadata caveat is gone.

* Added a run-level example key — "ci_run_url": "..."  # run-level context, so the "both scenario- and run-level" claim is now concrete.

nina-msft (dismissed):

docs/usage/results-and-reporting.md
```diff
@@ -121,4 +121,29 @@ exfil = report.population_summary(harm_category=HarmCategory.DATA_EXFILTRATION)
...
+    "evaluator_version": "response_contains@1.4.2",
+    "mitigation_ref": "SEC-1234",
+    "ci_run_url": "https://ci.example.com/runs/94821",  # run-level context
+})
+
+assert result, result.summary
+```
 
+!!! note
+    The framework automatically contributes a couple of keys to metadata — `test_name`, plus `harm_category` when the test carries a `@pytest.mark.harm` marker — so the persisted metadata will contain slightly more than what the snippet sets.
```
bashirpartovi: This is not correct, you should not mention internal bookkeeping fields because they do often change. For example, `test_name` is called `_pytest_test_name`. `harm_category` is not a metadata key at all, it is set on the `Result.harm_category` field and serialized as the top-level `harm_category` in the per-result projection. 

Maybe reword it as something like this:
```markdown
The framework adds internal, underscore-namespaced keys (`_pytest_test_name`, `_pytest_nodeid`, `_rampart_result_index`) to `result.metadata`, and separately records the harm category on the top-level `harm_category` field (not in metadata). 
```

It's best to advise teams to ignore `_pytest_*`/`_rampart_*` keys when diffing.

docs/usage/results-and-reporting.md
```diff
@@ -121,4 +121,29 @@ exfil = report.population_summary(harm_category=HarmCategory.DATA_EXFILTRATION)
 !!! note
     `ERROR` results are excluded from rate calculations. A transient infrastructure failure is not a safety finding.
 
+---
+
+## Portable Regression Receipt
+
+For CI gating, consider capturing a stable JSON artifact that your team can diff across runs. Put both scenario-level facts (what should stay stable across time) and run-level context (what was tested) in `result.metadata` to use as your regression receipt.
```
bashirpartovi: I think this like overclaims diffing capability.  The `JsonFileReportSink` file is written to a timestamped path and contains inherently non-deterministic fields (`duration_seconds`, `turns`, per-result durations). Only the user-chosen `result.metadata` subset is stable across runs. Presenting the sink output as a diffable receipt without this caveat sets teams up to diff noisy files. Maybe you can add one sentence clarifying that only the curated metadata keys are meant to be diffed, and suggest extracting that subset (e.g., via `jq`) rather than diffing the whole run report.

nina-msft: @bashirpartovi is this comment outdated? Previously this line suggested using the `JsonFileReportSink` as the regression receipt and I updated it to suggest that `result.metadata` be the regression receipt.

docs/usage/results-and-reporting.md
nina-msft: My bad on the incorrect key names! I agree with this comment, although I don't think `harm_category` is necessary to call out if not part of metadata since the point of this note is to illustrate that more will be added beyond what data the user provides.

docs/usage/results-and-reporting.md
bashirpartovi: ahh yes, I was commenting on the previous commit. But still this need modifications. We need to clarify this particular claim:
> .. stable JSON artifact that your team can diff across runs

This is only true about the metadata so adding a statement about that is needed.

nina-msft: Added statement to L149

docs/usage/results-and-reporting.md
nina-msft: Modified note to remove specific key names and advise to ignore the keys you mentioned

bashirpartovi (approved):

nina-msft:
@sumit1kr - I'm going to merge in this work, great job! Please do open another PR or issue if you have any further thoughts. Thank you for tackling this!

PR #73, opened 2026-06-04
[FEAT]: `pytest-xdist` support for distributed test execution

nina-msft:
### Summary

Adds first-class `pytest-xdist` support to RAMPART so attack/probe sessions can shard across worker processes and still produce a single coherent report. Worker processes serialize their results through xdist's `workeroutput` channel; the controller deserializes, merges, and writes the final report.

### What's included

**New: `_xdist.py`**
- Versioned schema (`rampart.xdist.v1`) for worker → controller payloads
- `_sanitize` JSON-safe coercion at the trust boundary with depth limit, byte-size cap, and ANSI-escape stripping on deserialize to defend against terminal injection from worker output
- `SizeLimitError` raised by `finalize_worker` when the serialized payload exceeds the configured cap; controller logs and continues
- Configurable cap via new pytest CLI option `--rampart-xdist-max-bytes` / ini option `rampart_xdist_max_bytes` (default 64 MiB), validated positive
- Sink resolution re-raises `KeyboardInterrupt`/`SystemExit`, catches `Exception`

**Trial-group aggregation across workers**
- Trial metadata is captured as data on `RampartSession` at collection time and shipped through the `rampart.xdist.v1` payload (`trial_specs`), so the controller aggregates trial groups from merged worker data instead of `session.items` — which is not reliably populated with trial clones on the controller at session finish. Without this, trial-group FAIL/PASS verdicts silently disappeared under xdist even though per-clone results were present.
- `trial_specs` is back-compatible: payloads without trials emit an empty list; malformed entries are skipped and non-finite thresholds are clamped on deserialize.

**Deterministic result ordering & sink resolution (review feedback)**
- Merged results sort by a stable key `(_pytest_nodeid, _rampart_result_index, _rampart_source_worker)` so aggregated reports are reproducible regardless of worker completion order. These bookkeeping keys are namespaced with leading underscores so they cannot collide with user-supplied result metadata. `_pytest_nodeid`/`_rampart_result_index` are assigned authoritatively (absorb-time enumerate; deserialize-time outer-key + enumerate); `_rampart_source_worker` is tagged on `handle_testnodedown`.
- New `pytest_rampart_sinks` hook (`_hooks.py`) is the authoritative sink source — keyed on hook-impl existence, not truthiness — and short-circuits the fixture path so projects defining both don't double-register. The `rampart_sinks` fixture remains the single-process fallback.
- Controller-side discovery now **unwraps a `@pytest.fixture`-wrapped parameterless `rampart_sinks`** to its underlying function (`_get_wrapped_function()` with `_fixture_function`/`__wrapped__` fallbacks) and calls it directly, restoring the documented session-fixture fallback under `-n`. Fixtures that depend on other fixtures still warn and point at the hook.
- Shared `strip_ansi` extracted to `rampart/common/text.py` (CSI/OSC/DCS/NF + C0/C1 stripping) and reused by both the plugin and the xdist boundary.

**Incomplete runs fail loudly (review feedback)**
- A lost or crashed xdist worker can leave every surviving test green. For a safety framework "green" must mean "ran and passed", so `pytest_sessionfinish` now forces a non-zero exit status (`OK → TESTS_FAILED`) whenever the merged session is incomplete. An already-failing status is left untouched so a stronger failure signal is never masked.
- Incomplete-run reasons are also surfaced as a terminal warning when workers crash or report partial data.

**Updated plugin hooks (`plugin.py`, `_session.py`)**
- `pytest_addoption` registers the new CLI/ini option; `pytest_addhooks` registers `pytest_rampart_sinks`
- Worker, controller, and non-xdist branches dispatched cleanly; worker branch wraps `finalize_worker` to surface size-limit truncation as a warning

**Reporting**
- `JsonFileReportSink` now projects `report.metadata`, so xdist run-mode info (worker_count, dist_mode, incomplete reasons) lands in the emitted JSON file

**Tests**
- 82 unit tests in `test_xdist.py` (serialization, sanitization, size limits, ANSI stripping, controller detection, sink resolution/unwrap, option parsing, ordering determinism, trial-spec round-trip/merge)
- 13 unit tests in `test_text.py` for shared `strip_ansi`
- `test_plugin.py` extended (60 tests) for sink-hook resolution, OSC sanitizing, the incomplete-warning path, and incomplete-run exit-status enforcement
- `test_json_file.py` extended (14 tests) for metadata projection
- Slow-marked unit `test_xdist_aggregation.py` (13 tests) exercising end-to-end aggregation across real worker subprocesses, including trial-group verdicts under both `--dist=loadgroup` and `--dist=load` (run with `-m slow`)

**Docs**
- New `xdist.md` user guide, incl. the `pytest_rampart_sinks` hook, sink precedence, `--dist=loadgroup` vs default `load` guidance, and durability limitations
- Configuration, CI-integration, pytest-integration, results-and-reporting, and API reference pages updated to describe the new option (under a dedicated "Pytest Options" section, not under env vars)
- Architecture page notes the controller/worker split

**Incidental**
- `onedrive.py`: file-level pyright directive suppressing the `Unknown*` cascade caused by `msgraph-sdk` shipping without type stubs. No behavior change.
- `_xdist.py`: `# ty: ignore[unresolved-attribute]` on the pytest-xdist runtime `config.workeroutput` attribute so `ty check` passes under the repo's pyright → ty migration.

### Security notes

The xdist boundary treats worker output as untrusted:
- All values pass through `_sanitize` → JSON-safe primitives only
- Strings are stripped of ANSI escape sequences on deserialize (including nested strings inside dicts/lists) to prevent terminal injection from a compromised/misbehaving worker
- Total payload size is capped to prevent controller memory exhaustion; truncation is recorded with a marker and raised as `SizeLimitError`
- Depth cap (`MAX_METADATA_DEPTH = 6`) prevents pathological nesting

### Validation
- `uv run ruff check rampart tests` → all checks pass; `ruff format --check` clean
- `uv run ty check` → all checks passed
- `uv run pytest tests/unit` → 610 passed
- `uv run pytest tests/unit/pytest_plugin/test_xdist_aggregation.py -m slow` → 13 passed (trial-group verdicts verified under `--dist=loadgroup` and `--dist=load`)
- `uv run --group docs mkdocs build --strict` → no cross-reference warnings (remaining output is Windows-only symlink noise from the font-download plugin; clean on CI/Linux)
- Ran `pytest -n 4 --dist=loadgroup` against the HelpDesk Bot example in `rampart-examples`: 1 report generated. Running with xdist took ~99.66s versus ~230.73s serial (yay for speed!)

## Breaking changes

None

## Checklist

- [x] Documentation updated


spencrr (commented):
.

bashirpartovi:
Great job @nina-msft, really a solid PR. The overall architecture makes sense to me, which is workers collect, the controller merges, and only the controller emits. I also like that the worker output path treats the boundary seriously rather than just passing objects around and hoping for the best. I feel like we should definitely keep that part even if some of the surrounding design changes.

The first thing I would fix before merge is ordering. `build_report()` sorts by `r.metadata.get("test_name", "")`, but `test_name` is derived from `node.nodeid.split("::")[-1]`, so it is only the trailing test name. That collides for same-named tests in different files, parametrized cases, and trial clones. Once two results have the same sort key, insertion order wins, and under xdist insertion order is basically worker completion order. So the output can still move around between runs. I think the fix is just to keep the full nodeid and sort on that. Maybe it is worth adding a small regression test that merges the same two worker payloads in reverse order and gets identical report output.

I am a little bit worried about the sink discovery piece. `discover_sinks_from_conftest()` / `_resolve_sink_candidate()` end up doing a partial version of pytest's fixture resolution which is to find a `rampart_sinks` attribute, unwrap fixture internals, inspect the signature, and call it only if it is parameterless. That works for the happy path, but if someone has a normal fixture like `rampart_sinks(tmp_path_factory)` or `rampart_sinks(my_config)`, _xdist_ mode skips that sink with a warning. If that fixture is their only sink configuration, the controller has no sinks to emit to, so the report they expected would be missing. I think that's a pretty silent problem, the tests could all be green, but the report never gets offloaded.

This would be cleaner as an explicit hook instead of fixture reflection. The key difference is that the existing `rampart_sinks` is a fixture, so it only gets resolved when something inside the fixture system asks for it. That works in single-process mode because the bootstrap fixture can call `request.getfixturevalue("rampart_sinks")`. In _xdist_ though, the controller is the process that needs to emit the final report, and the controller does not run tests or fixtures. That is why this PR has to scan conftest modules and manually unwrap the fixture function.

A hook would be different:

```python
def pytest_rampart_sinks(config):
    return [JsonFileReportSink(output_dir=Path(".report"))]
```

That would be a **hook** implementation, not another fixture. RAMPART would register the hook spec from its pytest plugin, and users would implement the hook from a root `conftest.py` or an installed plugin, the same way they implement other pytest hooks. Then the controller can call `config.pluginmanager.hook.pytest_rampart_sinks(config=config)` without running any fixtures. It still would not support fixture dependencies, but at least that limitation would be explicit instead of discovered through reflection.

The one thing to be careful about is having two sink configuration surfaces. If both `pytest_rampart_sinks` and the existing `rampart_sinks` fixture are present, the plugin should have a clear rule, probably something like "the hook is authoritative, the fixture is the legacy single-process fallback", and defining both generates a warning rather than emitting it twice. Without that precedence rule, this could get confusing quickly. With that rule documented, I would much rather support the hook as the xdist-compatible path than maintain a partial fixture resolver. I think for the longer term, it may be cleaner to deprecate the fixture-based sink registration entirely and make the hook the one supported sink API, once users have had time to migrate.

I think we need to put more throughts into whether to use `workeroutput` transport. I am not saying it is wrong, there is a good argument for it, since it is the native xdist channel and can support remote workers without requiring a shared filesystem. But it does mean results only get to the controller at worker shutdown. If a worker gets killed before `pytest_sessionfinish`, the run is correctly marked incomplete, but all results from that worker are gone, including tests that already finished. For small suites that may be fine. For long agent suites, that could be a lot of data to lose.

The size cap has a similar issue. If one worker payload goes over `--rampart-xdist-max-bytes`, we replace the whole payload with a truncation marker. So one huge transcript can make us lose the other results from that worker too. I do not think that has to block this PR, but I would call it out clearly in the docs. Longer term, a per-test stream or per-worker shard file would be more durable. `workeroutput` is simpler and works better for remote workers, so I mostly just want that tradeoff to be explicit.

One maintainability, there are now two serializers for roughly the same data. `_xdist.py` has `_serialize_result` / `_deserialize_result`, while `JsonFileReportSink` still has its own report serialization. Since the xdist path already has to round-trip `Result`, it might be worth pulling the shared bits into something like `rampart/reporting/serde.py` and making the JSON sink use it too; otherwise every future `Result` field has to be remembered in two places.

Two small notes

1. the `onedrive.py` pyright suppression looks unrelated to xdist, so I would probably split that out unless it is needed to get this PR green.
2. I am glad `incomplete` / `incomplete_reasons` are in the metadata, but I would think about surfacing incomplete runs somewhere CI users are likely to notice. The awkward case is all tests passing while the report is marked incomplete.

I like the direction. The two things I would most like to see tightened are the full-nodeid ordering and replacing the controller-side sink fixture scanning with an explicit hook or some similar explicit controller-safe sink registration path.

tests/unit/pytest_plugin/test_xdist_aggregation.py
spencrr, line 1: nit: move to unit tests

rampart/pytest_plugin/plugin.py
```diff
@@ -404,30 +498,33 @@ def rampart_sinks():
 
 def _aggregate_trial_results(
     *,
-    session: pytest.Session,
+    session: pytest.Session,  # noqa: ARG001  — kept for hook signature symmetry
     rampart_session: RampartSession,
 ) -> None:
```
spencrr, line 588: Should we just drop pytest session here. Is it true that we just stuff the results under the rampart key/rampart_session

rampart/pytest_plugin/plugin.py
```diff
@@ -489,11 +596,65 @@ def pytest_sessionfinish(
...
+        _record_xdist_metadata(session=session, rampart_session=rampart_session)
+        controller_sinks = discover_sinks_from_conftest(config=session.config)
+        if controller_sinks:
+            rampart_session.add_sinks(sinks=controller_sinks)
+        _emit_sinks(rampart_session=rampart_session)
+        return
+
     _aggregate_trial_results(session=session, rampart_session=rampart_session)
     _evaluate_gates(rampart_session=rampart_session)
     _emit_sinks(rampart_session=rampart_session)
```
spencrr, line 738: nit: possible pull out just the xdist parts

```suggestion
    _aggregate_trial_results(session=session, rampart_session=rampart_session)
    _evaluate_gates(rampart_session=rampart_session)
    if is_xdist_controller(config=session.config):
        _record_xdist_metadata(session=session, rampart_session=rampart_session)
        controller_sinks = discover_sinks_from_conftest(config=session.config)
        if controller_sinks:
            rampart_session.add_sinks(sinks=controller_sinks)
    _emit_sinks(rampart_session=rampart_session)
```

rampart/pytest_plugin/_xdist.py
```diff
@@ -0,0 +1,1166 @@
...
+    Args:
+        config (pytest.Config): The pytest configuration object.
+
+    Returns:
+        bool: True if running in the xdist controller process.
+    """
+    if is_xdist_worker(config=config):
+        return False
+    numprocesses = getattr(config.option, "numprocesses", None)
+    return numprocesses is not None and numprocesses != 0
```
spencrr: Possibly reconsider from ([xdist/plugin.py](https://github.com/pytest-dev/pytest-xdist/blob/master/src/xdist/plugin.py)) 

```python
def is_xdist_controller(request_or_session) -> bool:
    return (
        not is_xdist_worker(request_or_session)
        and request_or_session.config.option.dist != "no"
    )
```
This PR uses `numprocesses is not None and numprocesses != 0` instead. For `-n N` flows these agree (xdist's `pytest_cmdline_main` sets `dist = "load"` whenever `numprocesses` is truthy). They diverge for:

`pytest -d --tx=...` (the -d shortcut for --dist=load)
`pytest --dist=loadgroup --tx=2*popen` (any `--tx`-based setup without `-n`)
Any custom scheduler that sets `dist` without `-n`
In those cases, this PR's controller branch returns `False`, so `pytest_sessionfinish` falls into the non-xdist branch on the controller. The controller would then try to `_aggregate_trial_results` and `_emit_sinks` against its own (empty)` _results_by_nodeid` while workers still serialize into `workeroutput`. Therefore, the report would end up empty rather than incomplete.

docs/usage/xdist.md
```diff
@@ -0,0 +1,170 @@
...
+---
+
+## Trust Boundary & Security
+
+Worker payloads cross a process boundary via `execnet` and may contain attacker-controlled content (agent responses, payload text, evaluator rationale). RAMPART's serialization defends against:
+
+- **Arbitrary code execution** — strict JSON-safe primitives only; no `pickle`, `marshal`, or custom `__reduce__`.
+- **Schema drift** — payloads with missing or unknown schema versions are rejected fail-closed.
+- **Memory exhaustion** — worker payloads are capped at 64 MB by default.
+- **Terminal/log injection** — ANSI escape sequences are stripped from free-form text at the deserialization boundary.
```
spencrr, line 183: ai note:

**ANSI stripping doesn't cover OSC, and the security notes claim it does.**

`_ANSI_ESCAPE_RE = re.compile(r"\x1b\[[0-9;]*[A-Za-z]")` matches CSI only. It does not match:

- **OSC** (`\x1b]…BEL` or `\x1b]…ST`) — used for terminal hyperlinks (`\x1b]8;;https://attacker/\x1b\\text\x1b]8;;\x1b\\`) and window-title sets. Windows Terminal, iTerm2, modern GNOME Terminal, and Konsole all render hyperlinks. So an attacker-controlled response can produce a clickable link in `pytest`'s terminal output even though "ANSI was stripped".
- **DCS / SOS / PM / APC** (`\x1b[PXY^_]…ST`)
- **8-bit C1** (`\x9b…`) — less common but allowed by ECMA-48

The PR explicitly markets ANSI stripping as a defense against terminal injection from compromised workers, so this gap directly undermines a stated security property. A stricter pattern that handles the C1 introducers, or just `re.sub(r"[\x00-\x08\x0b-\x1f\x7f-\x9f]", "", text)` to strip all C0/C1 controls, would close it. Whichever you pick, also worth adding a couple of OSC-flavored test cases to `TestStripAnsi`.

rampart/pytest_plugin/_xdist.py
```diff
@@ -0,0 +1,1166 @@
...
+
+
+def _package_version() -> str:
+    """Return the installed rampart package version (best-effort)."""
+    try:
+        from importlib.metadata import version  # noqa: PLC0415
+
+        return version("rampart")
+    except Exception:  # noqa: BLE001
+        return "unknown"
```
spencrr: what case would this happen?

rampart/pytest_plugin/_xdist.py
```diff
@@ -0,0 +1,1166 @@
...
+    if trial_specs:
+        session.merge_trial_specs(trial_specs=trial_specs)
+    logger.info(
+        "Merged %d result group(s) from worker %s.",
+        len(results_by_nodeid),
+        worker_id_str,
+    )
+
+
+def discover_sinks_from_conftest(*, config: pytest.Config) -> list[ReportSink]:
```
spencrr, line 1069: ai find:

**`discover_sinks_from_conftest` is a parallel sink-discovery path with a silent capability gap.**

Two different mechanisms now resolve `rampart_sinks`:

| Process | Mechanism | Can satisfy fixture deps? |
| --- | --- | --- |
| Non-xdist / worker | `_rampart_sink_bootstrap` autouse fixture → `request.getfixturevalue("rampart_sinks")` | Yes |
| xdist controller | `discover_sinks_from_conftest` → plugin attribute scan + `inspect.signature()` | No (any param ≠ 0 ⇒ warn + skip) |

So `@pytest.fixture(scope="session") def rampart_sinks(tmp_path_factory): ...` works perfectly under `pytest -n0` and silently disappears under `pytest -n2`, leaving only a `logger.warning(...)`.

Consider:

- **Hook-based contributor API:** define a `pytest_rampart_sinks(config) -> list[ReportSink]` hookspec. Both controller and workers call it the same way; no fixture machinery dependency. Teams move their factory from a fixture into a hookimpl in conftest. The cost is asking users to update their conftest; the benefit is one unified path with no capability gap.


Also reaching into `candidate._get_wrapped_function()` is a private pytest API. The `__wrapped__` fallback is fine; the `_get_wrapped_function` call is at risk on any pytest minor bump.

spencrr (commented):
Hey Nina! I like:
- one (de)serialization spot in `_xdist.py` with sanitization and schme versioning.
- using `trial_specs` over worker channel so that `--dist-load` works correctly
- workers disable `_emit_sinks` and controller is the single source of truth

### Other small AI Finds
 
- **Worker-side `register_trial_spec` is dead work.** Workers run `pytest_collection_modifyitems`, register specs, ship them across, and never aggregate or evaluate. That's fine for correctness, but it means the spec is computed twice (once on the worker, once on the controller's own collection pass, which the merge then drops via `setdefault`). Worth a one-line comment in `pytest_collection_modifyitems` clarifying why we register on every process even though only the controller acts on it - otherwise future readers will be tempted to "optimize" the worker path and break the fallback.
- **`xdist_group` marker is added unconditionally to every trial clone.** Correct behavior (xdist sees it; pytest without xdist treats it as an unknown marker). If the repo ever turns on `--strict-markers` you'll need a conditional `addinivalue_line("markers", "xdist_group: ...")` somewhere; not a problem today.
- **`_session.py` and `_xdist.py` "documented deviations" notes.** The architecture doc was actually updated in this PR to describe `_session.py`/`_xdist.py` as first-class members of the plugin package - so the `Note: ... documented deviation from the architecture` blocks at the top of those modules are now stale and can come out. Minor.
- **Trial-spec aggregation runs on every controller `pytest_sessionfinish` call,** including when only one worker reported. That's fine for correctness; just observing that worker-result completeness isn't a precondition for trial aggregation, which is the right call - missing-worker should mark incomplete, not block the report.
- **`_emit_sinks` background-task path.** pre-existing - when an event loop is already running at session-finish (e.g., some pytest-asyncio configurations) the emission becomes fire-and-forget with no awaiter. Not changed by this PR, but the new xdist controller path is the first place I'd actually worry about this in production, because the controller's sink emission carries cross-worker data that didn't exist before. Worth at least filing as a follow-up.

nina-msft:
(In response to https://github.com/microsoft/RAMPART/pull/73#issuecomment-4673520297 above)

Thanks for the thorough read, @bashirpartovi — this shaped the last round of changes. Status on each point:

**Ordering (full nodeid):** Fixed. `build_report()` now sorts on `(nodeid, result_index, source_worker)` with the nodeid assigned authoritatively (not the trailing `test_name`), so same-named/parametrized/trial cases no longer collide. Added the reverse-order merge regression test you suggested — it merges the same two worker payloads in both orders and asserts identical report output.

**Sink discovery → explicit hook:** Done. Added a `pytest_rampart_sinks(config)` hookspec; the controller resolves it via `config.pluginmanager.hook` with no fixture machinery. Precedence is exactly what you proposed: **the hook is authoritative, the `rampart_sinks` fixture is the legacy single-process fallback**, and when a hook implementation exists the fixture path is skipped entirely (no double-registration). Documented in `xdist.md`. The fixture fallback still works on the controller for the common parameterless case (unwrapped to its underlying function), but anything with fixture deps now warns and points at the hook instead of silently disappearing.

**workeroutput durability & size-cap-drops-whole-payload:** Called out explicitly in a new "Durability limitations" section — a worker killed before `sessionfinish` loses its already-finished results, and an over-cap payload is dropped wholesale (worker marked incomplete). I agree the longer-term answer is a durable per-worker shard; I'm splitting that into a stacked follow-up PR on top of this one (incremental JSONL shards that survive a killed worker, with the cap applied per-record) so this PR stays focused.

**Two serializers:** I looked at extracting a shared module and decided to hold off for now. The two projections deliberately differ — the xdist one is a full-fidelity `Result` round-trip with sanitization + size handling, the JSON-sink one is the flatter public report shape — so unifying them today would couple things that legitimately diverge. I've added explicit "must not be naively merged" doc-comments on both and filed a follow-up to revisit once the shared surface is real. Open to doing it now if you feel strongly.

**onedrive pyright suppression:** Kept intentionally — it's a file-level pyright directive silencing the `Unknown*` cascade from `msgraph-sdk` shipping without type stubs (plus a `reportMissingImports` ignore on the import). No behavior change, and pyright/CI is red without it. Happy to pull it into its own PR if you'd prefer, but it's load-bearing for green CI on this branch.

**Surfacing incomplete runs:** Added a terminal warning at session end listing `incomplete_reasons`, so the "all tests green but report incomplete" case is visible in the pytest output, not just buried in metadata.

nina-msft:
(In response to https://github.com/microsoft/RAMPART/pull/73#pullrequestreview-4480545681 above)

Thanks @spencrr — addressed the small stuff:

- **Worker-side `register_trial_spec` dead work:** Added the clarifying comment in `pytest_collection_modifyitems` explaining why specs register on every process even though only the controller acts on them (so nobody "optimizes" the worker path and breaks the fallback).
- **Stale "documented deviation" notes:** Removed from `_session.py` and `_xdist.py` now that `architecture.md` lists them as first-class. (Kept the separate handler-factory note in `plugin.py` — different deviation.)
- **`_emit_sinks` fire-and-forget:** Agreed this is the first place it matters in production. Filed as a follow-up — out of scope for this PR since it's pre-existing.
- **`xdist_group` + `--strict-markers`:** Noted; not a problem today, will add the conditional `addinivalue_line` if/when we enable strict markers.
- **Trial aggregation per `sessionfinish`:** Right — worker completeness isn't a precondition; missing workers mark incomplete rather than block the report. Intentional.

(Replied inline on the individual threads with specifics.)

docs/usage/xdist.md
nina-msft: Good catch — the old regex was CSI-only. Replaced it with a shared `strip_ansi` in `rampart/common/text.py` that strips CSI **and** OSC, DCS/SOS/PM/APC, and 8-bit C1 introducers, then drops any residual C0/C1 control bytes (keeping only tab/newline/CR). Added OSC-flavored cases — the `\x1b]8;;…\x07` hyperlink, an ST-terminated window-title set, and an 8-bit C1 CSI — to the strip tests. The security-notes claim now matches the implementation.

rampart/pytest_plugin/_xdist.py
nina-msft: Reworked exactly as you flagged. `is_xdist_controller` now keys off the distribution, not the worker count: `not is_xdist_worker` and `get_dist_mode != "no"` and (`numprocesses` or `tx`). So `-d`, `--tx=…` without `-n`, and custom schedulers that set `dist` without `-n` all resolve to the controller branch instead of falling through to the empty single-process path; a bare `--dist` with no endpoints is still excluded.

rampart/pytest_plugin/_xdist.py
nina-msft: The hook (`pytest_rampart_sinks`) is now the unified, capability-explicit path you suggested — controller and workers resolve it the same way, no fixture machinery. The fixture fallback stays for single-process compat; on the controller it unwraps a parameterless fixture, but the `_get_wrapped_function()` call is wrapped in try/except with `_fixture_function` / `__wrapped__` fallbacks, so a pytest minor bump degrades to a warning (→ "use the hook") rather than crashing. A fixture with dependencies no longer silently disappears — it warns and points at the hook.

rampart/pytest_plugin/plugin.py
nina-msft: Done — hoisted `_aggregate_trial_results` / `_evaluate_gates` above the branch so the shared steps run once, and only `_record_xdist_metadata` + sink discovery are guarded by `is_xdist_controller`.

rampart/pytest_plugin/plugin.py
nina-msft: You're right, it was vestigial. Trial aggregation reads entirely from the merged `RampartSession` (`trial_specs`), never `session.items` — that's the whole reason the xdist path works. Dropped the unused `session` parameter (and its `noqa`).

rampart/pytest_plugin/_xdist.py
nina-msft: `importlib.metadata.version("rampart")` raises `PackageNotFoundError` when there's no installed distribution metadata for the package — e.g. running from a source checkout that was never `pip install`-ed, or a vendored/`PYTHONPATH` setup. The version is only a diagnostic stamp for spotting controller/worker version skew, so it must never crash a worker's serialization — hence the broad catch falling back to `"unknown"`.

tests/unit/pytest_plugin/test_xdist_aggregation.py
nina-msft: These spin up real xdist worker subprocesses end-to-end via the `pytester` fixture (`runpytest("-n", …)`), so they can't run in-process as unit tests — the cross-worker merge is exactly what they exercise. Keeping them under `tests/integration` for that reason; the pure serialization/sanitization logic is covered by the unit tests in `test_xdist.py`.

rampart/pytest_plugin/_xdist.py
spencrr: sorry, not following - how could we even get this far without RAMPART installed?

tests/unit/pytest_plugin/test_xdist_aggregation.py
spencrr: Pushing back here - these are not live LLM tests.

Maybe:
- `tests/unit/pytest_plugin/`
- `tests/functional/` or `tests/component/`
- ~~move `integration` to `e2e` and keep the `pytester` in `integration`~~ - dispreferred bc want to mirror naming with PyRIT

nina-msft: d'oh...thanks for the push back, i let ghcp run too rampant with the replies 🫡 

will move these into `test/unit/pytest_plugin` and mark the test with `@pytest.mark.slow`

rampart/pytest_plugin/_xdist.py
nina-msft: same as above - ur right & thanks for the push back! removed this entirely

spencrr:
> The one thing to be careful about is having two sink configuration surfaces. If both `pytest_rampart_sinks` and the existing `rampart_sinks` fixture are present, the plugin should have a clear rule, probably something like "the hook is authoritative, the fixture is the legacy single-process fallback", and defining both generates a warning rather than emitting it twice. Without that precedence rule, this could get confusing quickly. With that rule documented, I would much rather support the hook as the xdist-compatible path than maintain a partial fixture resolver. I think for the longer term, it may be cleaner to deprecate the fixture-based sink registration entirely and make the hook the one supported sink API, once users have had time to migrate.
> 
> I think we need to put more throughts into whether to use `workeroutput` transport. I am not saying it is wrong, there is a good argument for it, since it is the native xdist channel and can support remote workers without requiring a shared filesystem. But it does mean results only get to the controller at worker shutdown. If a worker gets killed before `pytest_sessionfinish`, the run is correctly marked incomplete, but all results from that worker are gone, including tests that already finished. For small suites that may be fine. For long agent suites, that could be a lot of data to lose.
> 
> The size cap has a similar issue. If one worker payload goes over `--rampart-xdist-max-bytes`, we replace the whole payload with a truncation marker. So one huge transcript can make us lose the other results from that worker too. I do not think that has to block this PR, but I would call it out clearly in the docs. Longer term, a per-test stream or per-worker shard file would be more durable. `workeroutput` is simpler and works better for remote workers, so I mostly just want that tradeoff to be explicit.
> 

@nina-msft ack-ing the main 3 points from @bashirpartovi here:

1. hook vs fixture
  - I would say that we can just deprecate the fixture for now for 0.2.0 and then remove in 0.3.0 or something similar. wdyt?
2. worker failure detection / surfacing to report
  - I think we can also tie into some of the xdist hooks here too
  - Copilot:
    > | Hook | When it fires | Use it to… | Runs on |
    > |---|---|---|---|
    > | **`pytest_testnodedown(node, error)`** | Any worker goes down (crash *or* graceful) | Detect worker death. `error` is non-`None`/truthy only for a real failure; `None` for clean shutdown. | Controller > only |
    > | **`pytest_handlecrashitem(crashitem, report, sched)`** | A worker died mid-test | Inspect/modify the synthetic failed report, or reschedule the test (`sched.mark_test_pending(crashitem)`) | Controller > only |
    > | **`pytest_runtest_logreport(report)`** | Every test result, *including* the synthetic crash failure | Your normal report aggregation — crashes already arrive here as `outcome="failed"` | Controller > (re-emitted from workers) |
3. truncation

Approving from my other review comments

spencrr (dismissed):

nina-msft:
@copilot resolve the merge conflicts in this pull request

rampart/pytest_plugin/plugin.py
```diff
@@ -607,16 +858,40 @@ def _write_trial_group_lines(
...
+def _write_incomplete_warning(
+    *,
+    terminalreporter: TerminalReporter,
+    rampart_session: RampartSession,
+) -> None:
+    """Write a prominent warning when the RAMPART run is incomplete.
+
+    Incomplete runs (e.g. a crashed or lost xdist worker) can still exit
+    zero with every surviving test passing, so the reasons are surfaced
+    here regardless of whether any results were collected.
```
bashirpartovi, line 900: This is not desirable behavior. For a safety framework, "green" should mean "ran and passed". Right now, this can silently drop a shard of tests, which should instead result in a non-zero exit status.

We should set a non-zero `session.exitstatus` in `pytest_sessionfinish` whenever `is_incomplete` is true.

rampart/pytest_plugin/_session.py
```diff
@@ -151,11 +211,16 @@ def absorb(self, *, node: pytest.Item, collector: ResultCollector) -> None:
...
             # Shallow copy is sufficient because we reconstruct all
             # mutable fields we modify (currently metadata and harm_category).
             result = copy.copy(original_result)
-            result.metadata = {**result.metadata, "test_name": test_name}
+            result.metadata = {
+                **result.metadata,
+                "test_name": test_name,
+                "nodeid": node.nodeid,
+                "result_index": result_index,
+            }
```
bashirpartovi, line 225: Maybe we should say `_pytest_nodeid` and `_rampart_result_index` to reduce the chances of overriding a similarly named variable in the metadata. We should actually re-think this but it is out of the scope of this PR. We are mixing user metadata (things they want to be emitted) with scheduling book-keeping.

rampart/pytest_plugin/plugin.py
nina-msft: Done in 1f383b4. `pytest_sessionfinish` now calls `_enforce_incomplete_exit_status`, which upgrades `session.exitstatus` to `TESTS_FAILED` whenever `rampart_session.is_incomplete` is true and the status is otherwise `OK`. It only upgrades `OK` → failed, so it never masks a stronger existing signal (e.g. `INTERRUPTED`). It runs on the controller and single-process paths — workers early-return in `pytest_sessionfinish` before it, so a worker's local incompleteness doesn't decide the controller's exit code. Added `TestIncompleteExitStatus` covering forced-fail, preserve-OK, and don't-mask-existing-failure.

rampart/pytest_plugin/_session.py
nina-msft: Done in 1f383b4. Renamed the bookkeeping keys with namespace prefixes as suggested: `_pytest_nodeid` and `_rampart_result_index`, plus `_pytest_test_name` and `_rampart_source_worker` for consistency (all four live in the same `absorb`/merge path and are internal-only — no docs or external consumers). Fully agree the deeper fix is separating user-emitted metadata from scheduling bookkeeping entirely; leaving that out of scope here as you noted.

spencrr (approved):
## Approving ✅ (with a few non-blocking follow-ups)

Really nice work on this - the worker-collect / controller-merge / controller-emit split is clean, and I appreciate that the `workeroutput` boundary is treated as untrusted rather than just passing objects across and hoping. I went through each resolved thread against the code on the branch and ran the suites locally:

- **169** targeted unit tests (`test_text`, `test_xdist`, `test_plugin`, `test_json_file`) - green
- **13** slow end-to-end aggregation tests under real worker subprocesses (`--dist=loadgroup` **and** `--dist=load`) - green

Looks like all the comments raised were addressed.

- Ordering now sorts on the full `_pytest_nodeid` (+ `_rampart_result_index`, `_rampart_source_worker`), assigned authoritatively - the three-way tie is structurally impossible, so it can't silently fall back to completion order. Reverse-merge regression test is there.
- `pytest_rampart_sinks` hook is authoritative, keyed on impl *existence* (not sink truthiness), fixture demoted to single-process fallback, no double-register. Controller unwrap degrades to a warning on a pytest bump rather than crashing.
- `strip_ansi` covers CSI/OSC/DCS/C1 + residual C0, is reused on both the transport and terminal paths, has no ReDoS exposure, and doesn't over-strip legitimate `[0m`-style text. Security-notes claim now matches reality.
- Incomplete run -> `OK -> TESTS_FAILED` only (never masks a stronger signal), workers early-return before enforcement. "Green means ran and passed" holds.

Approving so this isn't blocked. A couple of remaining small follow-ups:

1. **Vestigial clone attrs.** `clone._rampart_trial_base` / `_rampart_trial_index` are still set but aggregation reads `trial_specs` exclusively now - looks dead. Fine to fold into the future "separate user metadata from scheduling bookkeeping" cleanup. I think we can re look at using pytest's built in parameterization on a hidden trial index but not do the manual cloning.
2. **Determinism test could be sharper.** It uses distinct trailing names (`test_a` / `test_z`), so it proves order-independence but doesn't directly lock in the same-trailing-name-different-file collision that motivated the fix. A one-line case with a shared trailing name would nail it

Thanks, @nina-msft !

PR #86, opened 2026-06-17
[META]: Migrate pyright to ty

spencrr:
## Description

Migrate type checker from pyright to ty to align with PyRIT (microsoft/PyRIT#1319).

### Typing fixes prompted by ty

**`InjectionHandle` and `Session` now inherit `AbstractAsyncContextManager`** (rampart/core/injection.py, rampart/core/adapter.py). ty rejected `AsyncExitStack.enter_async_context(handle)` because a `Protocol` with `__aenter__ -> Self` doesn't unify with `AbstractAsyncContextManager[Handle, _ExitT_co]`'s covariant `_T_co` without an explicit subtype relationship.

```python
class InjectionHandle(AbstractAsyncContextManager["InjectionHandle", None], Protocol):
class Session(AbstractAsyncContextManager["Session"], Protocol):
```

Notes:
- typeshed declares `AbstractAsyncContextManager` as both `ABC` and `Protocol` ([stdlib/contextlib.pyi](https://github.com/python/typeshed/blob/main/stdlib/contextlib.pyi)) - it's on the type-checker allowlist, so a `Protocol` may legally inherit from it.
- Base order matters for C3 linearization: the ABC must precede `Protocol`.
- Generic args `[InjectionHandle, None]`: `_T_co` = `__aenter__` return; `_ExitT_co` = `None` because cleanup never suppresses exceptions (`closing`/`nullcontext` family, not `suppress`).
- `__aexit__` aligned with stdlib conventions: positional-only `/`, canonical `exc_value`/`traceback` names.

**`Attacks.xpia` and `coerce_driver` narrowing reordered** (rampart/attacks/__init__.py, rampart/drivers/_utils.py). ty cannot yet narrow element type through `isinstance(x, list)` over `list[A] | list[B]`. Reordering to check the protocol/scalar branch first eliminates the need for `cast` and `# type: ignore`.

**`ResponseContains` branch order tightened** (rampart/evaluators/response_contains.py). `callable()` doesn't narrow against a union containing `str` (every str is callable for ty's purposes here). Reordering to `re.Pattern` -> `str` -> `callable` and using `if found is True:` makes the dispatch unambiguous.

**`pytest_plugin/plugin.py`** (rampart/pytest_plugin/plugin.py): replaced a `register_default_handler_factory` lambda with a typed module-level `_default_handler_factory` so `list[ExecutionEventHandler]` aligns under list invariance. `pytest.Item` user-attribute writes now use `# ty: ignore[unresolved-attribute]`.

### Test changes

- tests/unit/attacks/test_xpia.py: `_mock_handle` and `_mock_evaluator` now use `spec=InjectionHandle` / `spec=Evaluator`. With the reordered narrowing in `Attacks.xpia` (`isinstance(inject, InjectionHandle)` first), bare `AsyncMock()` becomes ambiguous because runtime-checkable protocols treat any Mock as matching (every `hasattr` returns True). `spec=` makes `isinstance` deterministic.
- tests/unit/core/test_adapter.py: `@abstractmethod` added to a protocol-conformance test fixture so ty doesn't flag the `...` body.
- Test-only `# pyright: ignore` -> `# ty: ignore` (e.g., tests/unit/core/test_llm.py).

### Docs

Five contributing pages updated to reference ty: docs/contributing/index.md, development-setup.md, code-style.md, architecture.md, pull-requests.md.

## Breaking changes

None for users. Contributors will need to run `uv sync` to install ty (pyright is removed from the dev group) and update editor integrations - call out in development-setup.md.

## Checklist

- [x] Tests added or updated for changes - `tests/unit/attacks/test_xpia.py` mocks gain `spec=` for deterministic protocol `isinstance`; pyright ignore strings ported in `tests/unit/core/test_llm.py`, test_execution.py, `test_protocols.py`, test_types.py, etc.
- [x] Documentation updated - 5 contributing pages

---

### Verification

- `uv run ty check`: All checks passed
- `uv run ruff check . && uv run ruff format --check .`: clean
- `uv run pytest`: 491 passed, 5 skipped
- `uv run pre-commit run --all-files`: all hooks pass


rampart/core/adapter.py
```diff
@@ -19,7 +20,7 @@
 
 
 @runtime_checkable
-class Session(Protocol):
+class Session(AbstractAsyncContextManager["Session"], Protocol):
```
nina-msft, line 23: nit: Now that both protocols inherit `AbstractAsyncContextManager`, worth aligning their `__aexit__` signatures. `InjectionHandle.__aexit__` was updated to `exc_value`/`traceback` + positional-only `/`, but `Session.__aexit__` (lines 53-58 here) still uses `exc_val`/`exc_tb` with no `/`. ty accepts both — it's just the asymmetry that's confusing to a reader. Cosmetic.

tests/unit/attacks/test_xpia.py
```diff
@@ -5,8 +5,9 @@
 
 from unittest.mock import AsyncMock
 
-from rampart.attacks import Attacks
+from rampart.attacks import Attacks, InjectionHandle
```
nina-msft: nit: `InjectionHandle` isn't in `rampart.attacks.__all__` — it's importable here only because it became an incidental runtime import for the new `isinstance(inject, InjectionHandle)` check in `Attacks.xpia`. Prefer importing from its canonical home `rampart.core.injection` so the test doesn't depend on a re-export that could quietly disappear. The `spec=InjectionHandle` on the mock is the right call given that new runtime check.

tests/unit/core/test_adapter.py
```diff
@@ -40,6 +41,7 @@ class NotSession:
 class TestAgentAdapterProtocolCheck:
     def test_conforming_class_satisfies_protocol(self) -> None:
         class MyAdapter:
+            @abstractmethod
```
nina-msft: nit: `@abstractmethod` on a plain (non-ABC) class that's instantiated right below (`isinstance(MyAdapter(), AgentAdapter)`) is a runtime no-op and reads as the opposite of intent — this is meant to be a *conforming concrete* adapter, and the sibling `MySession` class above uses real bodies. It's only here to quiet ty about the `...` not returning `Session`. `raise NotImplementedError` would be clearer and works with any return type. Same thing in `tests/unit/core/test_protocols.py:66`. (Non-blocking.)

docs/contributing/code-style.md
```diff
@@ -7,12 +7,12 @@ RAMPART enforces a consistent code style through automated tooling and documente
 | Tool | Purpose | Config location |
 |------|---------|-----------------|
 | [Ruff](https://docs.astral.sh/ruff/) | Linting and formatting | `pyproject.toml` `[tool.ruff.*]` |
-| [Pyright](https://github.com/microsoft/pyright) | Static type checking (strict mode) | `pyproject.toml` `[tool.pyright]` |
+| [ty](https://github.com/astral-sh/ty) | Static type checking | `pyproject.toml` `[tool.ty]` |
```
nina-msft, line 10: The migration missed two Pyright references further down this same file: line 35 ("**Pyright** runs in **strict mode** targeting Python 3.11…") and line 160 ("…runs Ruff … and Pyright (strict type checking)…"). Both should move to ty to match this table. Minor related cleanup: `.gitignore` still ignores `pyrightconfig.json`.

pyproject.toml
```diff
@@ -136,5 +126,17 @@ convention = "google"
...
+python-version = "3.11"
+
+[tool.ty.src]
+include = ["rampart", "tests"]
+
+[[tool.ty.overrides]]
+include = ["tests/**"]
+
+[tool.ty.overrides.rules]
+invalid-argument-type = "ignore"
```
nina-msft: This is broader than the Pyright config it replaces, which only relaxed `reportPrivateUsage` for tests. Ignoring `invalid-argument-type` across all of `tests/**` means a genuine wrong-arg-type bug in any test goes unflagged — not just the intentional "pass a bad type and assert it raises" cases (which is why the targeted `# type: ignore[arg-type]` comments were dropped from test_llm_bridge / test_execution / test_plugin). If the goal is just enabling those negative tests, per-line `# ty: ignore[invalid-argument-type]` keeps the safety net for the rest of the suite. Is the blanket ignore intended?

rampart/evaluators/response_contains.py
```diff
@@ -42,18 +42,19 @@ async def evaluate_async(self, *, context: EvalContext) -> EvalResult:
...
             check_text = text if self._case_sensitive else text.lower()
             check_target = (
                 self._target if self._case_sensitive else self._target.lower()
             )
             found = check_target in check_text
+        elif callable(self._target):
+            found = self._target(text)
 
-        if found:
+        if found is True:
```
nina-msft: This `is True` is a behavioral regression. `_target` can be any callable, and a common pattern is `ResponseContains(lambda t: re.search(pat, t))` — that returns an `re.Match`, which is truthy but `Match is True` is `False`, so it now reports `NOT_DETECTED` where the old `if found:` reported `DETECTED`. For a safety detector that's a silent false-negative (test passes while the agent actually leaked).

ty doesn't require this — after the isinstance reorder `found` is statically `bool`, so plain `if found:` type-checks fine. Suggest reverting just this line to `if found:`. The isinstance reordering above it is a genuine improvement — keep that.

docs/contributing/code-style.md
spencrr: accepting docs references, leaving .gitignore as default from python template

pyproject.toml
spencrr: Fair enough, just a lot of noise with all the mocks -- AI finding below:

**The core issue:** When you write `MagicMock(spec=MyClass)`, the static type of the expression is still `MagicMock` (more precisely `unittest.mock.MagicMock`). The `spec=` argument is a *runtime* construct — it configures the mock's attribute/method surface and makes `isinstance()` checks pass at runtime — but it carries no static type information. So ty correctly sees a `MagicMock` being passed where `MyClass` is expected, and `MagicMock` is not assignable to `MyClass`, hence `error[invalid-argument-type]`.

This is **not a ty bug** — every static type checker (mypy, pyright, ty) behaves this way, because the `spec` argument's value isn't reflected in the return type. There is an open feature request to make checkers understand `spec` ([python/mypy#5217](https://github.com/python/mypy/issues/5217) and related), but no checker resolves the spec type today.

## Answer / Deliverable

You have four real options, ordered from what I'd recommend most to least for production code.

### 1. Annotate the variable as the target type (best for readability) — **High confidence**

```python
from unittest.mock import MagicMock

my_mock_obj: MyClass = MagicMock(spec=MyClass)
call_my_func(my_mock_obj)
```

The declared type `MyClass` is what flows into `call_my_func`, so no error. This is clean and keeps the intent obvious. The tradeoff: after this line, the checker treats `my_mock_obj` as a plain `MyClass`, so mock-specific attributes (`.assert_called_once()`, `.return_value`, etc.) will themselves raise errors. If your test needs to call those *and* pass the object as `MyClass`, prefer option 2.

### 2. `cast()` at the call site (best when you also need the mock API) — **High confidence**

```python
from unittest.mock import MagicMock
from typing import cast

my_mock_obj = MagicMock(spec=MyClass)   # stays MagicMock — mock API works
call_my_func(cast(MyClass, my_mock_obj))  # only this expression is MyClass
```

`cast` is erased at runtime (zero cost) and narrowly scopes the type assertion to the call, leaving `my_mock_obj` usable as a mock for assertions. This is the most common pattern in well-typed test suites.

### 3. `create_autospec` + `cast` (best runtime fidelity)

```python
from unittest.mock import create_autospec
from typing import cast

my_mock_obj = create_autospec(MyClass)
call_my_func(cast(MyClass, my_mock_obj))
```

`create_autospec` enforces the real signatures of `MyClass`'s methods at runtime (catching bad calls your tests make), but statically it's still `MagicMock`, so you still need the `cast`. Use this when you want stronger runtime guarantees, not to silence the checker by itself.

### 4. Targeted suppression (last resort)

```python
call_my_func(MagicMock(spec=MyClass))  # ty: ignore[invalid-argument-type]
```

Only if you can't restructure. Prefer a scoped `ty: ignore[invalid-argument-type]` over a blanket ignore so other real errors on the line still surface.

---

**Recommendation:** Default to **option 2 (`cast`)** in test code — it's explicit, zero-runtime-cost, doesn't hide the mock's own API, and survives refactors. Reach for **option 1 (annotation)** when the object is *only* passed through and you don't touch the mock API afterward.

**One caveat to flag:** all four approaches assert a shape the checker can't verify. If `MyClass`'s real interface changes, none of these will catch a now-incorrect mock — that's the inherent cost of mocking against a static checker, and the reason `create_autospec` (option 3) is worth the extra fidelity for mocks that stand in for complex collaborators.

Sources: [unittest.mock — `spec`/`create_autospec` docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock), [mypy discussion on `spec`-aware typing (python/mypy#5217)](https://github.com/python/mypy/issues/5217).

rampart/evaluators/response_contains.py
nina-msft: Optional follow-up (non-blocking): the revert to `if found:` is the right fix, but the gap that let `if found is True:` slip through is still untested. The callable tests in `tests/unit/evaluators/test_response_contains.py` (`TestResponseContainsCallable`) only use `lambda text: "secret" in text`, which returns a real `bool` — so they'd pass under *both* `if found:` and the buggy `if found is True:`.

Consider adding one regression test with a callable that returns a truthy non-`bool`, mirroring real usage like `ResponseContains(lambda t: re.search(pattern, t))` (which yields an `re.Match`, not `True`):

```python
async def test_callable_returning_truthy_non_bool_detects(self) -> None:
    evaluator = ResponseContains(lambda text: re.search(r"sec\w+", text))
    result = await evaluator.evaluate_async(context=_ctx("secret"))
    assert result.outcome is EvalOutcome.DETECTED
```

That locks the behavior so this can't silently regress again.

nina-msft (approved):

rampart/evaluators/response_contains.py
spencrr: reverted to `found` is truth-y / `found.__bool__` returns `True`. I don't think it needs to be a reg test :)

PR #95, opened 2026-06-22
[META]: Generate package metadata dynamically for PyPI

spencrr:
## Description

Move RAMPART package metadata generation to Hatchling so release artifacts can produce PyPI-safe metadata without requiring manual README or version edits.

### Dynamic README metadata

The source `README.md` now keeps repository image links relative, while package builds generate PyPI README metadata with raw GitHub image URLs. This keeps the README readable in GitHub and makes images render reliably on PyPI. The build-time README generation uses Hatch’s dynamic metadata hook pattern ([Hatch dynamic metadata](https://hatch.pypa.io/latest/how-to/config/dynamic-metadata/), [custom metadata hooks](https://hatch.pypa.io/latest/plugins/metadata-hook/custom/)).

### Dynamic version metadata

Package versions are now derived from Git tags through Hatch VCS instead of a static `project.version` field. Release builds get their version from tags like `vX.Y.Z`; development builds after a tag infer a development version. Local version suffixes such as `+g<sha>` are omitted for public-package compatibility via `local_scheme = "no-local-version"` ([hatch-vcs version source](https://github.com/ofek/hatch-vcs#version-source), [setuptools-scm local schemes](https://setuptools-scm.readthedocs.io/en/latest/extending/#setuptools_scmlocal_scheme)).

The release process docs were updated to describe tag-derived versions, relative README image links, and the requirement that release tags be reachable from `main` history for expected development-version inference.

### Package data inclusion

The setuptools-specific prompt inclusion configuration was removed as part of the Hatchling migration:

```toml
[tool.setuptools.package-data]
rampart = [
    "drivers/prompts/*.yaml",
    "evaluators/prompts/*.yaml",
]
```

Hatchling does not use `tool.setuptools.package-data`. Instead, it uses Hatch’s file-selection model. With:

```toml
[tool.hatch.build.targets.wheel]
packages = ["rampart"]
```

the wheel target explicitly selects the `rampart/` package directory. Hatch documents `packages` as semantically equivalent to `only-include`, with package paths rewritten as needed. Selected directories are traversed by default, subject to normal VCS ignore and `exclude` behavior.

I verified the built wheel directly and confirmed the prompt YAML files are still present:

```text
rampart/drivers/prompts/llm_driver_system_prompt.yaml
rampart/evaluators/prompts/llm_judge.yaml
```

This preserves the previous package data behavior while removing configuration that only applied to the old setuptools backend. References: [Hatch file selection](https://hatch.pypa.io/latest/config/build/#file-selection), [Hatch `packages`](https://hatch.pypa.io/latest/config/build/#packages), and [wheel default file selection](https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection).

## Breaking changes

None. This changes package build metadata generation only; runtime APIs and package contents are unchanged.

## Checklist

- [X] Tests added or updated for changes - existing unit tests cover runtime behavior; package build and wheel metadata inspection verify the packaging changes.
- [X] Documentation updated - release process docs now cover dynamic versioning and README metadata behavior.

---

### Verification

- `uv lock --check`: passes
- `uv run ruff check scripts/hatch_build.py`: passes
- `uv run --with build python -m build`: passes
- `uv run pytest tests/unit -q`: 489 passed
- Wheel metadata inspection:
  - Generated by Hatchling
  - Version inferred dynamically without a local `+g<sha>` suffix, instead `.dev{commits_since_tag}`
  - README image path rewritten from relative source path to a raw GitHub URL

### DIFF
--- a/docs/contributing/release-process.md
+++ b/docs/contributing/release-process.md
@@ -34,14 +34,20 @@ If you find functionality to remove, merge the removal PR to `main` before proce
 
 ## 4. Update the Version
 
-### pyproject.toml
-Set the version in `pyproject.toml` to the version established in step 2.
+### Git tag
+RAMPART derives package versions from Git tags using Hatch VCS and setuptools-scm. No `pyproject.toml` version bump is required for a release. The release version is determined by the `vx.y.z` tag pushed in step 5.
 
 ```toml
-[project]
-name = "RAMPART"
-version = "x.y.z"
+[tool.hatch.version]
+source = "vcs"
+
+[tool.hatch.version.raw-options]
+local_scheme = "no-local-version"
 ```
+
+The `no-local-version` setting omits local version suffixes such as `+g<sha>` because PyPI does not support them for upstream releases. See the [setuptools-scm local scheme documentation](https://setuptools-scm.readthedocs.io/en/latest/extending/#setuptools_scmlocal_scheme) for details.
+
+For development builds on `main`, the release tag must be reachable from `main` history for Hatch VCS to infer the next development version from that tag. If the release branch contains commits beyond `main`, merge or cherry-pick those release commits back to `main` after publishing.
 
 ### Update README File
 The README file is published to PyPI and also needs to be updated so the links work properly. _Note: There may not be any links to update, but it is good practice to check in case our README changes._
@@ -149,7 +155,7 @@ If successful, the URL `https://pypi.org/project/rampart/x.y.z/` will return the
 
 After the release is on PyPI, open a PR to `main` containing only:
 
-- In line with PyPA [versioning guidance](https://packaging.python.org/en/latest/discussions/versioning/), bump the version in `pyproject.toml` to the next development version (e.g., `x.y.(z+1).dev0` or `x.(y+1).0.dev0`, depending on the next planned release).
+- Any follow-up documentation or metadata updates needed after the release. Do not bump the package version in `pyproject.toml`; once `main` has commits after the release tag, Hatch VCS will infer the next development version automatically.
 - Replace any references to the previous release version in the codebase with the new released version (without `.dev0`) where applicable (e.g., installation docs that pin to the latest tag).
 
 Open this PR from a branch separate from your `releases/vx.y.z` branch.
@@ -216,10 +222,10 @@ A patch release (e.g., `0.2.0` → `0.2.1`) ships a targeted fix — typically a
 
     Resolve any conflicts manually. Patch-sized fixes typically apply cleanly.
 
-3. **Bump the version** in `pyproject.toml` to the new patch version. Also update any version-pinned links in `README.md`.
+3. **Update release-specific references** as needed. Do not bump the package version in `pyproject.toml`; the patch version comes from the `vx.y.z` tag. Also update any version-pinned links in `README.md`.
 
     ```bash
-    git commit -am "Bump version to x.y.z"
+    git commit -am "Prepare x.y.z release"
     ```
 
 4. **Push and tag**:
--- a/docs/contributing/release-process.md
+++ b/docs/contributing/release-process.md
@@ -54,9 +54,9 @@ The README file is published to PyPI and also needs to be updated so the links w
 
 Keep README image links relative when they point to files in this repository, e.g., `docs/images/RAMPART.svg`. During package builds, `scripts/hatch_build.py` generates the PyPI README metadata and rewrites those image paths to raw GitHub URLs with the release version.
 
-Replace any other “main” links like “doc/index.md” with “raw” links that have the correct version number, i.e., “https://raw.githubusercontent.com/microsoft/RAMPART/releases/vx.y.z/docs/index.md”.
+Replace any other "main" links like "doc/index.md" with "raw" links that have the correct version number, i.e., "https://raw.githubusercontent.com/microsoft/RAMPART/releases/vx.y.z/docs/index.md".
 
-For directories, update using the “tree” link, e.g., “https://github.com/microsoft/RAMPART/tree/releases/vx.y.z/docs/usage"
+For directories, update using the "tree" link, e.g., "https://github.com/microsoft/RAMPART/tree/releases/vx.y.z/docs/usage"
 
 This is required for the release branch because PyPI does not pick up other files besides the README, which results in local links breaking.
 
--- a/docs/contributing/release-process.md
+++ b/docs/contributing/release-process.md
@@ -46,9 +46,9 @@ version = "x.y.z"
 ### Update README File
 The README file is published to PyPI and also needs to be updated so the links work properly. _Note: There may not be any links to update, but it is good practice to check in case our README changes._
 
-Replace all “main” links like “doc/index.md” with “raw” links that have the correct version number, i.e., “https://raw.githubusercontent.com/microsoft/RAMPART/releases/vx.y.z/docs/index.md”.
+Keep README image links relative when they point to files in this repository, e.g., `docs/images/RAMPART.svg`. During package builds, `scripts/hatch_build.py` generates the PyPI README metadata and rewrites those image paths to raw GitHub URLs with the release version.
 
-For images, update using the “raw” link, e.g., “https://raw.githubusercontent.com/microsoft/RAMPART/releases/vx.y.z/docs/images/RAMPART.png”.
+Replace any other “main” links like “doc/index.md” with “raw” links that have the correct version number, i.e., “https://raw.githubusercontent.com/microsoft/RAMPART/releases/vx.y.z/docs/index.md”.
 
 For directories, update using the “tree” link, e.g., “https://github.com/microsoft/RAMPART/tree/releases/vx.y.z/docs/usage"
 
--- a/docs/contributing/release-process.md
+++ b/docs/contributing/release-process.md
@@ -32,10 +32,19 @@ If you are incrementing the minor version, search the codebase for the new minor
 
 If you find functionality to remove, merge the removal PR to `main` before proceeding.
 
-## 4. Update the Version
+## 4. Prepare Release Metadata
 
-### Git tag
-RAMPART derives package versions from Git tags using Hatch VCS and setuptools-scm. No `pyproject.toml` version bump is required for a release. The release version is determined by the `vx.y.z` tag pushed in step 5.
+Before tagging the release:
+
+- Do not add or update a version in `pyproject.toml`. The release version comes from the `vx.y.z` tag created in step 5.
+- Review `README.md` for repository-relative links that need to work on PyPI.
+- Keep image links under `docs/images/` relative. During package builds, `scripts/hatch_build.py` rewrites them to raw GitHub URLs pinned to the release version.
+- For other repository-relative links, use absolute `https://github.com/microsoft/RAMPART/...` URLs on `main`, or extend `scripts/hatch_build.py` to rewrite them at build time.
+- Merge any required README or metadata changes to `main` before continuing to step 5.
+
+### Why the Tag Must Be on `main`
+
+RAMPART derives package versions from Git tags using Hatch VCS and setuptools-scm:
 
 ```toml
 [tool.hatch.version]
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -63,14 +64,17 @@ Issues = "https://github.com/microsoft/RAMPART/issues"
...
+[tool.hatch.metadata.hooks.custom]
+path = "scripts/hatch_build.py"
 
-[tool.setuptools.package-data]
-rampart = [
-    "drivers/prompts/*.yaml",
-    "evaluators/prompts/*.yaml",
-]
+[tool.hatch.version]
+source = "vcs"
```
nina-msft, line 72: AI find

---

Deriving the version from VCS assumes a release tag is reachable from the branch being built. Our release process puts `vX.Y.Z` on a `releases/vX.Y` branch and doesn't merge it to `main`, so that assumption doesn't hold here:

* `git merge-base --is-ancestor v0.1.0 origin/main` → fails (tag isn't an ancestor of main)
* `git describe --tags HEAD` → "No tags can describe HEAD"
* `uv build --wheel` → `rampart-0.1.dev75`, and `git rev-list --count HEAD` is also `75` → counting from repo root, no tag found
* `packaging.Version("0.1.dev75") < Version("0.1.0")` → `True`

So `main` builds sort before the shipped `0.1.0`, and stay in `0.1.devN` even after future releases. Note `fallback_version` won't help — it only applies when there's no `.git` at all (e.g. building from an sdist); a `main` checkout has git metadata but no reachable tag, so setuptools-scm uses its no-tag counting instead.

Fix: make release tags reachable from `main` — either tag a commit that's on `main`, or add a release step that merges the tagged release commit back into `main` (a real merge, not cherry-pick). I tested this: with the tag reachable, `git describe` →` v0.1.0-35-g…` and the build produces `0.1.1.dev35`, which correctly sorts after 0.1.0. Please confirm the intended flow lands the tag on main before this merges.

docs/contributing/release-process.md
```diff
@@ -34,23 +34,29 @@ If you find functionality to remove, merge the removal PR to `main` before proce
...
+[tool.hatch.version]
+source = "vcs"
+
+[tool.hatch.version.raw-options]
+local_scheme = "no-local-version"
 ```
 
+The `no-local-version` setting omits local version suffixes such as `+g<sha>` because PyPI does not support them for upstream releases. See the [setuptools-scm local scheme documentation](https://setuptools-scm.readthedocs.io/en/latest/extending/#setuptools_scmlocal_scheme) for details.
+
+For development builds on `main`, the release tag must be reachable from `main` history for Hatch VCS to infer the next development version from that tag. If the release branch contains commits beyond `main`, merge or cherry-pick those release commits back to `main` after publishing.
```
nina-msft: AI find

---

"merge or cherry-pick those release commits back to main" — cherry-pick creates new commits with new SHAs, so the tagged commit stays unreachable and git describe still won't find it. Only merging the tagged commit (or tagging something already on main) satisfies setuptools-scm. This is the same root cause as the version blocker above.

Also, this guidance sits in §4 (before the tag is even created in §5), and neither §5 nor §6 adds a step that brings the tag onto main — which is why v0.1.0 isn't on main today. Please add an explicit "merge the release tag into main" step and drop the cherry-pick wording.

scripts/hatch_build.py
```diff
@@ -0,0 +1,81 @@
...
+_RELATIVE_HTML_IMAGE_URL_PATTERNS = (
+    re.compile(r'(<img\b[^>]*\bsrc=")(?:\./)?(docs/images/[^"]+)(")'),
+    re.compile(r"(<img\b[^>]*\bsrc=')(?:\./)?(docs/images/[^']+)(')"),
+)
+_RELATIVE_MARKDOWN_IMAGE_URL_PATTERN = re.compile(
+    r"(!\[[^\]]*\]\()(?:\./)?(docs/images/[^)]+)(\))",
+)
+
+
+def _readme_ref(version: str) -> str:
```
nina-msft, line 33: Consider adding unit tests for these:

`_readme_ref`, `_raw_image_url`, and `_render_readme` are pure and deterministic but untested, and a regex regression (an image silently not rewritten, or a mangled README) would ship without a signal. They're trivially importable — worth a small tests/ module covering: dev version → main ref; release version → vX.Y.Z ref; HTML src="docs/images/…" and ./docs/images/…; markdown `alt`; the already-absolute main URL patterns; and no-image passthrough.

scripts/hatch_build.py
```diff
@@ -0,0 +1,81 @@
...
+"""Hatchling metadata hooks for RAMPART package builds."""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from hatchling.metadata.plugin.interface import MetadataHookInterface
+
+_GITHUB_IMAGE_URL_PATTERNS = (
```
nina-msft, line 15: nit:

Only docs/images/… paths are rewritten, so a relative image elsewhere (e.g. assets/…) would pass through and 404 on PyPI. The markdown pattern (docs/images/[^)]+) would also pull an optional title into the URL for` a`. Neither bites the current README — just flagging in case image usage grows. A one-line comment pinning the docs/images/ assumption would be enough.

nina-msft (commented):
Overall, this deviates from our alignment with PyRIT's release process. Do we think if it is sucessful with RAMPART we will also update PyRIT's process to automate these parts?

pyproject.toml
spencrr: Good catch, @nina-msft:

Reproduced it: a normal `main` checkout builds `0.1.devN` (currently `0.1.dev88`), which sorts before `0.1.0`, and it won't self-correct since each merge just bumps N. Agreed that `fallback_version` doesn't apply here.

Root cause is what you said: `v0.1.0` lives on `releases/v0.1`, one commit off the merge-base, so it's never an ancestor of main. Two options:

A) Backfill now:
```
git merge -s ours v0.1.0 -m "Make v0.1.0 reachable from main"
```
`-s ours` leaves main's tree untouched (that commit only tweaks a README badge main already has) and just records the ancestry, so builds become `0.1.1.devN`. Downside: it has to be a real two-parent merge (a squash drops the second parent and does nothing), so it breaks our squash-only linear history.

B) Wait for 0.2.0 and tag a commit that's on main, then branch `releases/v0.2` from there instead of tagging the branch. The build hook handles the README rewrite now, so the tag can live directly on main and history stays linear. Downside: main builds stay `0.1.devN` until 0.2.0 ships, which is fine unless we start publishing dev builds.

Either way we need the process fix, which is really your closing question: today we tag the release branch, so the tag never reaches main. I'll also fix the release doc, since the "merge or cherry-pick" line is wrong (cherry-pick makes a new SHA, so reachability still fails).

Leaning B plus the doc fix, assuming no interim dev builds. If we do publish them, let's do A now and still fix the process so 0.2.0 doesn't repeat this.

docs/contributing/release-process.md
spencrr: Fixed in 986fe34. Dropped the "merge or cherry-pick" wording (cherry-pick makes a new SHA, so the tag stays unreachable and `git describe` still misses it) and moved tagging into §5 so it happens on a commit that's on `main`, with the release branch cut from that tag. §7 now also re-tags the `main` commit rather than the release branch when fixing issues found during testing. So there's an explicit "tag lands on main" step now.

scripts/hatch_build.py
spencrr: Done in 881bcd3. Added a comment pinning the `docs/images/` assumption, and tightened the Markdown pattern so the URL group stops at whitespace, preserving an optional title like `alt` instead of folding it into the rewritten URL. Non-`docs/images/` paths (e.g. `assets/`) are intentionally left out of scope for now, and the comment notes they need to be absolute URLs. Regression is covered by `test_titled_markdown_image_preserves_title`.

scripts/hatch_build.py
spencrr: Yeah, it may be a bit much, but I added them in 02eff4b: `tests/scripts/test_hatch_build.py` covers `_readme_ref` (dev/local - `main`, release - `vX.Y.Z`), `_raw_image_url`, and `_render_readme` across relative and `./` HTML, Markdown, titled Markdown, the already-absolute `main` URL patterns, and no-image passthrough. They run in a dedicated workflow (ebb0dac) that triggers on `scripts/` and `tests/scripts/` changes.

pyproject.toml
spencrr: Went with option B. The release-process doc now lands the tag on `main`: §5 tags a commit that's on `main` and cuts `releases/vX.Y` from that tag, and §7 re-tags the `main` commit (not the branch) if we fix something during release testing (986fe34). So the intended flow does put the tag on `main` going forward. `0.1.x` builds stay `0.1.devN` until `0.2.0` ships, which is harmless as long as we aren't publishing dev builds. If you'd rather correct `0.1` right now instead of waiting, we can think about doing the merge but leaning against pretty strongly.

spencrr:
> Overall, this deviates from our alignment with PyRIT's release process. Do we think if it is sucessful with RAMPART we will also update PyRIT's process to automate these parts?

💯

docs/contributing/release-process.md
```diff
@@ -34,36 +34,48 @@ If you find functionality to remove, merge the removal PR to `main` before proce
 
 ## 4. Update the Version
```
nina-msft: This whole section is just informational now instead of action for maintainer right? There's a lot of context here so I'd just make it clear what the action step is in Section 4 and what is just more info.

docs/contributing/release-process.md
```diff
@@ -34,36 +34,48 @@ If you find functionality to remove, merge the removal PR to `main` before proce
...
+Tag the release on `main` first, then cut the release branch from that tag. Tagging `main` rather than the release branch is what keeps the tag reachable from `main`, so development builds version correctly (see the [Git tag](#git-tag) note in step 4).
 
-Commit your changes to a release branch and push the tag:
+Confirm any release-prep changes have already merged to `main`, then:
 
 ```bash
-git checkout -b releases/vx.y.z
-git commit -am "release vx.y.z"
-git push origin releases/vx.y.z
+git checkout main
```
nina-msft, line 55: This change makes sense to me :)

docs/contributing/release-process.md
spencrr: Yep I suppose most of 4 is just context for why git tagging main in 5.

docs/contributing/release-process.md
```diff
@@ -216,10 +228,10 @@ A patch release (e.g., `0.2.0` → `0.2.1`) ships a targeted fix — typically a
 
     Resolve any conflicts manually. Patch-sized fixes typically apply cleanly.
 
-3. **Bump the version** in `pyproject.toml` to the new patch version. Also update any version-pinned links in `README.md`.
+3. **Update release-specific references** as needed. Do not bump the package version in `pyproject.toml`; the patch version comes from the `vx.y.z` tag. Also update any version-pinned links in `README.md`.
```
nina-msft: Can you include an example of what a release-specific reference would be? I think we can remove the "do not bump the package version....tag" sentence because we are explaining rationale for that earlier in this doc.

spencrr: Anything like "deprecated at v0.2.0" or something like that but for now I think we can tighten!

docs/contributing/release-process.md
spencrr: Updated Section 4 to lead with the maintainer actions, with the Hatch VCS and tag-reachability rationale moved into a separate informational subsection. The actual tagging commands remain in Section 5.

docs/contributing/release-process.md
spencrr: Tightened this. Added examples of release-specific references, such as "Fixed in v0.2.1" or README links pinned to a release tag, and removed the repeated package-version sentence.

docs/contributing/release-process.md
```diff
@@ -32,50 +32,32 @@ If you are incrementing the minor version, search the codebase for the new minor
...
-git tag -a vx.y.z -m "vx.y.z release"
-git push origin vx.y.z
-
-# Cut the release branch from the tagged commit, for release-only
-# artifacts and future patch releases.
-git checkout -b releases/vx.y.z vx.y.z
-git push origin releases/vx.y.z
+# Create the tag and release branch.
+git tag -a vx.y.0 -m "vx.y.0 release"
+git checkout -b releases/vx.y vx.y.0
```
nina-msft, line 60: Ok so the branch name drops the `.0` suffix but the tag keeps it? Why not just keep everything consistent with `v.x.y.0`?

nina-msft (approved):
One question about branch naming, but overall docs read more clearly. Thank you!

docs/contributing/release-process.md
spencrr: branches are for minor versions -> we create a new minor branch when we have a new minor version (along with setting patch to 0). In the case we have patches before the next minor release, we will create a new tag (at patch level)!

PR #103, opened 2026-07-07
[TEST]: silence pytest-asyncio deprecation warning in xdist pytester tests

spencrr:
## Description

The `tests/unit/pytest_plugin/test_xdist_aggregation.py` suite spawns child pytest sessions via `pytester`. Each child session is configuration-isolated from the repository's `pyproject.toml`, so pytest-asyncio read an empty `asyncio_default_fixture_loop_scope` via `config.getini(...)` and emitted a `PytestDeprecationWarning` once per subprocess run [(14 warnings across the suite)](https://github.com/microsoft/RAMPART/actions/runs/28828897791/job/85498298080#step:6:64).

This PR resolves the warnings by writing a minimal ini file into each child project root that mirrors the parent project's asyncio configuration (`asyncio_mode = auto`, `asyncio_default_fixture_loop_scope = session`), so the option is supplied through the same channel pytest-asyncio actually reads.

To keep it DRY, the child conftest and ini creation are consolidated into a single `configured_pytester` fixture that every test now depends on, removing the scattered per-test `pytester.makeconftest(...)` boilerplate (and the now-unnecessary `# pyright: ignore` comments on `makepyfile(...)`).

Result: `13 passed`, warning-free.

### Alternatives considered

- **Set the option from the child `conftest.py` `pytest_configure` hook** — e.g. `config.option.asyncio_default_fixture_loop_scope = "function"` or `config.addinivalue_line("asyncio_default_fixture_loop_scope", "function")`. Rejected: pytest-asyncio reads the value via `config.getini("asyncio_default_fixture_loop_scope")`, so the `config.option` namespace is never consulted, and `addinivalue_line` only appends to *line-list* ini options (e.g. `markers`, `filterwarnings`), not a scalar string option. Neither reaches `getini(...)`, so the warning persisted.
- **Pass `-o asyncio_default_fixture_loop_scope=session` to every `runpytest(...)`** — Works, but repeats the override across ~14 call sites; the ini file centralizes it in one place.
- **Disable the plugin per child run via `-p no:asyncio`** — Silences the warning (the child sessions are all synchronous), but diverges from the real project configuration and sidesteps configuring the very option these tests should exercise.
- **Ignore it via `filterwarnings`** — A band-aid that hides the symptom rather than setting the option, and would also mask the warning if it ever surfaced legitimately elsewhere.

## Breaking changes

None.

## Checklist

- [X] Tests added or updated for changes
- [ ] Documentation updated

### DIFF
--- a/tests/unit/pytest_plugin/test_xdist_aggregation.py
+++ b/tests/unit/pytest_plugin/test_xdist_aggregation.py
@@ -439,7 +439,8 @@ class TestXdistTrialAggregation:
 
 class TestXdistMetadata:
     def test_report_includes_xdist_metadata(
-        self, configured_pytester: Pytester
+        self,
+        configured_pytester: Pytester,
     ) -> None:
         _setup_simple_tests(configured_pytester)
         configured_pytester.runpytest("-p", "no:cacheprovider", "-n", "2")
@@ -475,7 +476,8 @@ class TestXdistMetadata:
 
 class TestCollectOnly:
     def test_collect_only_does_not_emit_reports(
-        self, configured_pytester: Pytester
+        self,
+        configured_pytester: Pytester,
     ) -> None:
         _setup_simple_tests(configured_pytester)
         configured_pytester.runpytest("-p", "no:cacheprovider", "--collect-only")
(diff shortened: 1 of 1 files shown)


nina-msft (approved):
Thanks for addressing this!

PR #105, opened 2026-07-07
[FEAT]: Deprecate the `rampart_sinks` fixture in favor of the `pytest_rampart_sinks` hook

nina-msft:
## Summary

Follow-up to #73. Now that the `pytest_rampart_sinks` hook is the supported
sink-registration API, this PR begins deprecating the legacy `rampart_sinks`
fixture.

Resolving the fixture now emits a `DeprecationWarning` (removal target: `0.3.0`)
in both fixture-consumption paths:

- **Single-process** — `_rampart_sink_bootstrap`, after the fixture resolves.
- **xdist controller** — `_resolve_sink_candidate`, when it unwraps a genuine
  `@pytest.fixture`-form `rampart_sinks` (not module-level lists or plain
  callables).

The hook stays authoritative and the fixture remains a single-process fallback,
so nothing breaks for existing users — they just get a migration nudge.

## Changes

- **New** `rampart/common/deprecation.py` — `emit_deprecation_warning(...)`,
  mirroring PyRIT's helper (qualified-name resolution + `DeprecationWarning` at
  `stacklevel=3`).
- Wire it into the `plugin.py` and `_xdist.py` fixture-resolution paths.
- Docs: deprecation admonitions in `usage/pytest-integration.md` and
  `usage/xdist.md`.
- Tests: `tests/unit/common/test_deprecation.py` (unit) and a
  `TestSinkFixtureDeprecation` class in the slow aggregation suite that drives
  real subprocess pytest runs via `pytester` — single-process warn,
  xdist-controller warn, and list-form no-warn — replacing the previous
  `MagicMock`-based warn/no-warn tests.

## Breaking changes

None. The fixture continues to work as a single-process fallback; it only emits
a `DeprecationWarning` now.

## Validation

- `uv run ruff check .` → all checks pass; `ruff format --check` clean.
- `uv run ty check` → all checks passed.
- `uv run pytest tests/unit/common/test_deprecation.py tests/unit/pytest_plugin/test_xdist.py tests/unit/pytest_plugin/test_plugin.py` → 145 passed.
- `uv run pytest tests/unit/pytest_plugin/test_xdist_aggregation.py -m slow` → 16 passed (includes the `pytester`-based `TestSinkFixtureDeprecation` class).

## Notes

- `removed_in="0.3.0"` is the removal milestone for the fixture — deprecated in
  the `0.1` line, kept through `0.2`, removed in `0.3` (two minor versions, per
  `docs/contributing/release-process.md`).
- Implements the first of the post-merge deprecation follow-ups raised by
  @bashirpartovi.

## Checklist

- [x] Documentation updated


docs/usage/pytest-integration.md
```diff
@@ -77,6 +77,12 @@ async def test_with_threshold(adapter):
 
 ### `rampart_sinks`
 
+!!! warning "Deprecated"
+    The `rampart_sinks` fixture is deprecated and will be removed in `0.3.0`.
```
nina-msft, line 81: I picked our current version +0.2.0 but lmk if we want a different strategy

spencrr (commented):
Looks good! First quick pass:
- print_deprecation_message -> log deprecation warning maybe?
- teats have a lot of mocks -> let’s use pytester and capture warnings that way?

nina-msft:
@spencrr I just updated the test name to be `emit_deprecation_warning` before seeing your comment so that is already addressed :D I'll do another pass through the tests shortly!

tests/unit/pytest_plugin/test_xdist_aggregation.py
```diff
@@ -45,6 +45,20 @@ def rampart_sinks():
...
+
+from rampart.reporting import JsonFileReportSink
+
+
+_OUT_DIR = Path("rampart_reports").absolute()
+_OUT_DIR.mkdir(parents=True, exist_ok=True)
+Path("rampart_report_dir.txt").write_text(str(_OUT_DIR))
+
+rampart_sinks = [JsonFileReportSink(output_dir=_OUT_DIR)]
+"""
```
spencrr: do we have to define the hook here as a func or plain list?

bashirpartovi (commented):
This is great @nina-msft . However, there are a few places in the documentation that need to be updated:

-  Quickstart doc, we need to update [Step 4:](https://microsoft.github.io/RAMPART/getting-started/quickstart/#step-4-add-reporting)
- `concepts/overview.md` describes the feature as "Report sinks: Structured output via the rampart_sinks fixture" see [pytest-integration](https://microsoft.github.io/RAMPART/concepts/overview/#pytest-integration)
- `ci-integration.md` shows a full `@pytest.fixture … def rampart_sinks()` code sample as the main example (it does have a downstream !!! tip pointing at the hook for xdist, so it's the least bad of the 3, but the headline example is still the fixture).

tests/unit/pytest_plugin/test_xdist_aggregation.py
nina-msft: Good flag — this pushed me to reconsider whether the list form needed a `pytester` subprocess test at all, and it didn't.

I've removed `_LIST_CONFTEST` and `test_controller_list_form_does_not_warn_under_xdist`, and replaced them with two fast in-process tests in `test_xdist.py::TestSinkDeprecationWarning`:
- `test_fixture_form_emits_deprecation_warning` — fixture form warns
- `test_list_form_does_not_emit_deprecation_warning` — list form stays silent

Reasoning: that subprocess test's discovery + single-report assertions already overlapped `test_finds_list_rampart_sinks` and `test_xdist_emits_single_consolidated_report`. Its only unique coverage was "the list form emits no deprecation warning" — and that's a trivial in-process branch (`_resolve_sink_candidate` returns early for lists, before `emit_deprecation_warning`). Asserting it directly is milliseconds instead of a ~30–60s `slow` xdist run, and it removes the list-form conftest that was the source of the confusion here. As a bonus, the fixture-warns direction now has an explicit unit assertion too — previously it was only incidental.

Pushed in 8fccfca.

bashirpartovi (approved):
LGTM

spencrr (approved):

PR #113, opened 2026-07-10
[FEAT]: Durable xdist shard transport primitives (no wiring)

nina-msft:
## Description

First PR in a **stack of two** that adds durable result transport for pytest-xdist, closing two gaps in the current in-memory `workeroutput` transport:

- **Durability** — a worker killed mid-run currently loses *all* of its finished results (nothing is written until a clean `pytest_sessionfinish`).
- **Granular size cap** — a single oversized transcript currently replaces a whole worker's payload with a truncation marker, discarding every other result from that worker.

This PR lands only the **primitives** for the fix. Every symbol is a pure addition to `rampart/pytest_plugin/_xdist.py` with **no call sites** — existing behavior is completely unchanged. It establishes the vocabulary the wiring PR consumes, so it can be reviewed small and green on its own.

**What's added**

| Symbol | Purpose |
| --- | --- |
| `ShardWriter` | Append-only, flush-per-result JSONL writer. A worker's finished results hit disk immediately, so they survive an abrupt worker death (e.g. SIGKILL). An oversized record is written as a marker line naming its `nodeid`/`index` instead of discarding the whole shard. |
| `read_worker_shard` | Controller-side recovery. Parses a worker shard line by line, tolerates a truncated/partial final line, and surfaces corrupt/oversized records as `DroppedRecord` entries so the caller can mark the run incomplete. `nodeid`/`index` are taken **authoritatively from the record envelope**, mirroring `deserialize_worker_data`'s existing trust posture (never trusting values embedded in the serialized result body). |
| `shard_eligible` | Gates shard transport to runs where every `--tx` gateway is a local `popen` worker (shared filesystem). Remote (`ssh=`/`socket=`) or proxied (`via=`) topologies are ineligible and will fall back to `workeroutput` in PR B. |
| `SHARD_DIR_KEY` | Stash key constant for the controller→worker shard-directory channel used in PR B. |

The full-fidelity `Result` round-trip reuses the existing `_serialize_result`/`_deserialize_result` pair, so shards carry the same fidelity as the current transport.

## 🔜 Coming in PR B — wire durable transport (the behavior change)
See draft here: https://github.com/nina-msft/RAMPART/pull/3

PR A is inert on its own. The follow-up (`dev/nina-msft/xdist-durable-transport`, stacked on this branch) turns the primitives on:

- Controller creates the shard directory and hands it to workers via a `pytest_configure_node` channel (`SHARD_DIR_KEY`).
- Workers append each result to their shard as it completes and stamp `"transport": "shard"`.
- **The atomic flip:** `workeroutput` is slimmed to a completeness *sentinel* (result bodies removed) **and** `handle_testnodedown` switches to reading results from shards — these two changes must land together, which is exactly why they're held out of this PR.
- Mandatory remote/non-eligible **fallback to `workeroutput`** (so single-process and remote runs are unaffected).
- Slow `pytester` tests proving the wiring: worker-kill recovery, oversized-record isolation, crash-item-absent, and clean-run parity; plus the `docs/usage/xdist.md` update.

Splitting here keeps the "flip the source of truth" as one reviewable semantic change and keeps this PR a zero-risk, additive foundation.

## Breaking changes

None. No existing code path calls any of the new symbols; behavior is identical to `main`.

## Checklist

- [ ] Documentation updated

### DIFF
--- a/rampart/pytest_plugin/_xdist.py
+++ b/rampart/pytest_plugin/_xdist.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 import json
 import logging
 import math
+from dataclasses import dataclass
 from datetime import datetime
 from typing import TYPE_CHECKING, Any, cast
 
@@ -48,6 +49,7 @@ from rampart.reporting.sink import ReportSink
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from pathlib import Path
 
     import pytest
 
@@ -62,6 +64,8 @@ DEFAULT_SIZE_LIMIT_BYTES: int = 64 * 1024 * 1024
 MAX_METADATA_DEPTH: int = 6
 
 _TRUNCATED_MARKER: str = "rampart_truncated"
+
+SHARD_DIR_KEY: str = "rampart_shard_dir"
 
 
 class WorkerOutputError(Exception):
@@ -139,6 +143,43 @@ def get_worker_count(*, config: pytest.Config) -> int:
     """
     numprocesses = getattr(config.option, "numprocesses", 0)
     return int(numprocesses) if numprocesses else 0
+
+
+def _is_local_popen_spec(*, spec: str) -> bool:
+    """Return whether one ``--tx`` gateway spec targets a local popen worker.
+
+    Args:
+        spec (str): An xdist execnet gateway spec (e.g. ``"popen"`` or
+            ``"popen//python=python3.11"``).
+
+    Returns:
+        bool: True when the spec launches a same-host popen worker with
+            no proxy routing (no ``ssh=``/``socket=``/``via=``).
+    """
+    head = spec.split("//", 1)[0].strip()
+    return head == "popen" and "via=" not in spec
+
+
+def shard_eligible(*, config: pytest.Config) -> bool:
+    """Return whether durable shard transport is usable for this run.
+
+    Shard transport requires every worker to share the controller's
+    filesystem, which holds exactly when all ``--tx`` gateway specs are
+    local ``popen`` workers. Remote (``ssh=``/``socket=``) or proxied
+    (``via=``) gateways may not see the controller's shard directory, so
+    those runs fall back to the in-memory ``workeroutput`` transport.
+
+    Args:
+        config (pytest.Config): The pytest configuration object.
+
+    Returns:
+        bool: True when at least one gateway is configured and every
+            configured gateway is a local popen worker.
+    """
+    tx = getattr(config.option, "tx", None)
+    if not tx:
+        return False
+    return all(_is_local_popen_spec(spec=str(spec)) for spec in tx)
 
 
 def _size_limit(*, config: pytest.Config) -> int:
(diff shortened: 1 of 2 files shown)


spencrr (commented):
@nina-msft We spoke offline about the filesystem vs in-mem serialized+streamed approaches. What about the following?

```mermaid
sequenceDiagram
    participant T as Test / RAMPART execution
    participant W as xdist worker
    participant E as execnet
    participant C as xdist controller
    participant S as RAMPART sink

    T->>W: execution emits Result(s)
    W->>W: ResultCollectionHandler records in item collector
    W->>W: call-phase pytest_runtest_makereport
    W->>W: serialize Result envelope(s)<br/>onto namespaced TestReport attribute
    W->>E: xdist testreport event
    E->>C: serialized TestReport
    C->>C: reconstruct TestReport
    C->>C: pytest_runtest_logreport
    C->>C: validate + merge Result envelope(s)
    Note over C: repeat incrementally per item

    W-->>C: workeroutput completion sentinel<br/>(expected count + trial specs)
    C->>C: verify streamed count / mark incomplete
    C->>S: emit one unified report
```

xdist already streams every setup/call/teardown `TestReport` from workers as tests execute, including for `popen`, ssh+socket, and proxied gateways. Its worker calls `pytest_report_to_serializable()` and sends a `testreport` event. The controller reconstructs the report and re-emits `pytest_runtest_logreport`.

Pytest’s default serializer copies the complete `report.__dict__`, and `TestReport` restores unknown fields through `**extra`. This is also an established plugin pattern:

- `pytest-html` attaches structured data as `report.extras`.
- `pytest-rerunfailures` attaches `report.rerun`.
- Core subtests uses explicit report serialization because it introduces a distinct `SubtestReport` type.
- Session-wide plugins such as `pytest-cov` use `workeroutput`, which is appropriate for data that has no natural per-item report.

We can namespace all of this under a private field such as `report._rampart_results`, rather than `user_properties`. Like you mentioned, `user_properties` is intentionally consumed by reporters such as JUnit XML, so putting complete prompts, responses, tool calls, and transcripts there could leak or substantially inflate JUnit output.

The small `workeroutput` sentinel would retain the valuable completeness properties of the current implementation:

- Expected result count.
- Trial specs.
- Schema/transport version.
- Clean worker completion.

That gives us these failure semantics:

| Failure | Streamed design |
|---|---|
| Worker dies after previous items completed | Previously received item reports remain merged |
| Worker dies after call report, before teardown/session finish | Current item remains merged |
| Worker dies during the test body, before its call report | Current item is missing; xdist emits a crash failure and missing sentinel marks the RAMPART run incomplete |
| Remote worker/channel dies | Previously received reports survive; unlike shards, this works for remote gateways |
| One result exceeds the cap | Send a dropped-result marker; mark run incomplete without dropping sibling results |
| Worker exits cleanly | Sentinel verifies streamed count and contributes trial specs |
| Controller dies | In-memory aggregate is lost; the proposed shards do not currently implement controller restart/recovery either |

There is one implementation precondition: today `_rampart_collect` absorbs results during fixture teardown, after the call report has already been created. A call-report transport needs access to the collector snapshot before that point. I view the collector-lifecycle change as separate work and am not proposing that it be folded into this PR.

## Why I prefer evaluating this before committing to shards

The shard approach introduces a second transport with a materially different support envelope:

- It works only for local `popen` workers; remote workers fall back to the existing worker-session batch and retain its larger failure window.
- It writes attacker-influenced prompts, responses, tool calls, and transcripts to cleartext temporary files.
- It requires directory provisioning, controller→worker path propagation, writer lifetime management, cleanup, retention options, reader recovery, and transport dispatch.
- Controller `SIGKILL`/OOM can leave sensitive shard files behind indefinitely; in-process cleanup cannot prevent that.
- The configured limit is per record, so total disk use is effectively unbounded across a large run.

The shard primitives are carefully implemented and well tested. My concern is whether the filesystem is necessary at all, rather than the code quality of the primitives.

The filesystem approach does have one important potential advantage: large transcripts do not pass through and queue on the controller’s execnet report channel. Streaming large payloads could increase serialization cost, channel backpressure, controller RSS, and scheduling latency. I think we should measure that rather than assume it dominates.

Could we prototype the report-streamed path and compare it using representative p50/p95/max RAMPART result sizes under:

- `-n 4` local workers.
- A remote or socket gateway.
- Worker termination after completed items.
- One oversized result among normal siblings.
- Thousands of small results.
- Representative 1 MiB, 10 MiB, and maximum-size transcripts.

The comparison should measure wall time, controller/worker RSS, bytes transferred, disk footprint, and recovered-result completeness.

If full-result streaming proves too expensive, a compact hybrid may be preferable:

- Stream the safety status, scores, trial identity, and compact metadata in the `TestReport`.
- Put unusually large transcripts behind an explicit artifact abstraction and stream an artifact reference.
- Keep completeness and gating data on the controller for every topology.

bashirpartovi:
@spencrr I like that idea but it's definitely not cheap to implement. I think this is worth prototyping before we decide which direction to take. We can mock the agent side and generate synthetic Result payloads, but still run real xdist workers so we're testing the actual report channel.

I'd probably start with the collector timing since results are currently absorbed during teardown. Then we can try a few sample payload sizes and see whether streaming the full results holds up or whether we need the hybrid.

@nina-msft does this sound good to you too?

nina-msft:
[@bashirpartovi](https://github.com/bashirpartovi) [@spencrr](https://github.com/spencrr) — prototype-first was the right call, so I ran it. I used Bashir's recipe: mock the agent, generate synthetic `Result` payloads, but drive **real** xdist workers so we're testing the actual report channel. Started with collector timing like you suggested.

**Decision: go with the streamed approach and drop the filesystem shards — and stream the full `Result`, no hybrid needed.** The prototype backs this up on every axis we cared about.

Before I'd trust that call, three things had to hold, and the prototype checked each one: **(1)** a `Result` bolted onto a report actually survives the worker→controller trip; **(2)** it's reachable at the moment we'd grab it, given how the collector is timed; and **(3)** the one we actually worried about — big payloads on the channel don't cost the controller memory or latency. One section each, then why that settles shards vs. streaming.

All three checks ran from a minimal, RAMPART-free rig (a bare conftest + trivial tests, no `import rampart`) so the numbers reflect the transport itself, not our heavy import cost. I disabled the installed plugin with `-p no:rampart` and neutralized the repo's pyproject options with `-o addopts=""` so nothing leaked into the runs. Everything ran in the worktree venv via `uv run --no-sync`, pytest 9.0.3 / xdist 3.8.0 / execnet 2.1.2 / py3.12.

### 1. Streaming works the way Spencer described

I stamped an unknown attribute onto `report._rampart_results` at the worker's call-phase `pytest_runtest_makereport`, then read it back on the controller in `pytest_runtest_logreport` under `-n 2` popen. It survived serialize → execnet → deserialize and showed up intact — the full nested payload (schema / nodeid / index + str/int/bool/null/list/dict) for every test. So the `report.__dict__`-copy + `TestReport(**extra)` path holds. The one rule: payloads have to be plain JSON types — which is exactly what `_serialize_result` already produces.

### 2. The collector-timing problem is real but solvable

I modeled the real `_rampart_collect` pattern (activate a ContextVar collector, `yield`, absorb after the body) with a call-phase makereport hookwrapper. At makereport the results are already there and reachable via both the ContextVar and `item.stash`, *before* the teardown absorb runs — so a streamed transport can grab them at the right moment. Async bodies recording from `create_task`/`gather` worked too, since asyncio copies the context into child tasks. As you said Spencer, the collector-lifecycle change is separate work — I'll do it as its own PR before starting the streaming work outlined in the design.

### 3. Streaming big payloads does NOT blow up the controller

This was your one open question, Spencer — whether large payloads on the channel cost us in memory or latency. Answer: no.

The controller has to hold every result in memory to build the final report — that's true today (`_serialize_report` groups over all results) and it's just as true for shards (`read_worker_shard` reads every record back into memory before merging). So the real question isn't "does streaming use memory," it's "does streaming use *more* memory than that floor."

**How I ran it.** The rig has two halves: one *generates* a controlled load, the other *measures* what the controller pays to receive it.

Generating the load — two files:

- *Mock agent* — a call-phase `pytest_runtest_makereport` hookwrapper stamps `report._rampart_results` onto each report. The payload is a small dict (`safe`, `summary`) plus one dominant `transcript` field of exactly `SPIKE_PAYLOAD_BYTES` bytes, mirroring a real Result whose size is dominated by transcript text. No LLM, no evaluation — just a synthetic payload whose size I control.
- *Trivial tests* — `test_stream` parametrized over `range(SPIKE_N)`, each body just `assert True`. So the only thing moving through the system is the payload — no agent work, no test logic to muddy the numbers.

Measuring the cost — three controller-side hooks:

- *Retention* — a controller-only `pytest_runtest_logreport` hook (the controller has no `PYTEST_XDIST_WORKER` env var; workers do) appends every payload it receives to a list and never drops it — reproducing the real controller holding all results in memory until the final merged report.
- *Completeness* — `received` is that list's length at the end. If it equals N, nothing was dropped in transit.
- *Sampling* — a daemon thread wakes every 30 ms and records both controller RSS and whole-tree RSS (controller + all four workers, via psutil `children(recursive=True)`), keeping baseline and peak. Wall time is monotonic from `pytest_configure` to `pytest_sessionfinish`, where the rig prints one `PROBE_RESULT` line.

I swept three payload sizes against real `-n 4` popen workers:

```
SPIKE_PAYLOAD_BYTES=<size> SPIKE_N=<count> \
  uv run --no-sync pytest <probe_dir> -n 4 -p no:rampart -s -q -o addopts=""
```

Results — *streamed* = total bytes sent (N × payload); *RSS growth* = controller memory gained over baseline; *overhead* = how much more than those raw bytes (the streaming-specific cost above the floor):

| Payload | N | streamed | controller RSS growth | overhead | delivered | wall |
|---------|---|----------|----------------------|----------|-----------|------|
| 10 KiB | 500 | 4.9 MiB | 11.4 MiB | (too small to measure) | 500/500 | 44s |
| 1 MiB | 200 | 200 MiB | 204 MiB | ~2% | 200/200 | 37s |
| 10 MiB | 100 | 1000 MiB | 1013 MiB | ~1% | 100/100 | 12s |

Reading it: every run delivered all N results, and the controller's memory growth tracked the streamed bytes almost exactly — ~1–2% over, i.e. streaming adds essentially nothing on top of the data the controller must hold anyway. At the 10 MiB scale the whole-tree peak hit ~1.3 GiB (controller + 4 workers) with ~1 GiB pushed through the channel: every result arrived, no crash, no stall. (Wall time tracks test *count*, not payload size — that's why the 500-test run was slowest.)

So, to answer your fork directly, Bashir: full-result streaming holds up, so **we don't need the hybrid.** You scoped it as a fallback for if streaming got too expensive — it didn't.

That's all three questions green: it survives the trip, it's reachable in time, and it's cheap on the channel. Which collapses the whole shards-vs-streaming debate into a one-way comparison.

### Why drop the shards

Shards don't save us any controller memory — the controller ends up holding the same data either way. What they *add* is everything you already flagged: popen-only (no remote workers), attacker-influenced transcripts written to disk in cleartext, disk use that's effectively unbounded across a run, and files left behind if the controller gets SIGKILLed. The one thing shards could have won on — keeping big payloads off the channel — turned out not to matter, because the channel handles them fine. So streaming wins on durability, topology, security, and performance. There's no axis left where shards come out ahead.

### What I did NOT test (being upfront)

This was a quick spike, not the full harness. I hit the performance-critical cases (`-n 4`, thousands of small results, 1 MiB and 10 MiB transcripts) and confirmed nothing gets dropped. I did **not** run the failure-mode rows from your matrix — remote/socket gateway, killing a worker mid-item, one oversized result among normal ones. Those are cases where streaming is *better* than shards on paper (it's the only design that survives a remote channel dying), so they don't change the decision — but they belong in the real prototype's test suite, and I'll cover them there.

One thing I won't oversell: ~1% means streaming doesn't *add* cost, not that memory is free — at 1 GiB of transcripts the controller really does hold ~1 GiB (shards would too). If total volume ever gets that big, your hybrid is the lever (stream metadata, artifact-ref the big transcripts) — a future scaling question, not a reason to keep shards.

### Next step

I'll build the real prototype, starting with the collector-lifecycle change as its own PR.

Any objections before I start? Also open to talk about reprioritization with other work items on the board if this will take a bit to implement and we prefer to start with something else.

PR #114, opened 2026-07-14
[FIX]: Fail pytest when a RAMPART trial gate fails

behnam-o:
Propagates failed aggregate trial gates to pytest’s exit status.

- Sets `session.exitstatus` to `TESTS_FAILED` when any trial group fails.
- Preserves existing non-zero exit statuses.
- Keeps individual trial items passing while the aggregate threshold determines the process exit code.

### DIFF
--- a/rampart/pytest_plugin/plugin.py
+++ b/rampart/pytest_plugin/plugin.py
@@ -676,6 +676,28 @@ def _enforce_incomplete_exit_status(
         )
 
 
+def _enforce_trial_gate_exit_status(
+    *,
+    session: pytest.Session,
+    rampart_session: RampartSession,
+) -> None:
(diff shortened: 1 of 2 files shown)


spencrr (commented):
@behnam-o @nina-msft Is this similar to https://github.com/microsoft/RAMPART/pull/73#discussion_r3508733534 but also includes upgrading trial failures to exit status as well?

spencrr (approved):
Thanks @behnam-o! While we still have trial clones, I think this fits with my mental model of RAMPART: any trial group that did not pass (`UNSAFE > 0 || SAFE / ALL < SAFE_THRESHOLD`) should raise pytest exit status to failed. @bashirpartovi thoughts here?

behnam-o:
closing this in favor of the work the will re-layer the trials into the execution logic

PR #115, opened 2026-07-14
FEAT Handle trial assertions through aggregate threshold evaluation

behnam-o:
Trial clone assertion failures were reported as individual pytest failures, bypassing the trial group’s aggregate threshold.

This change intercepts assertion failures for `@pytest.mark.trial` calls, records missing outcomes as `UNDETERMINED`, and marks clones as passed so the aggregate threshold determines the final result. Successful trials without explicit results are recorded as `SAFE`. Non-trial assertions and non-assertion exceptions retain normal pytest behavior. Tests cover threshold outcomes, result deduplication, and failure preservation.

### DIFF
--- a/rampart/pytest_plugin/plugin.py
+++ b/rampart/pytest_plugin/plugin.py
@@ -68,6 +68,7 @@ __all__ = [
     "pytest_addoption",
     "pytest_collection_modifyitems",
     "pytest_configure",
+    "pytest_runtest_makereport",
     "pytest_sessionfinish",
     "pytest_terminal_summary",
     "pytest_testnodedown",
@@ -76,6 +77,7 @@ __all__ = [
 
 _rampart_key = pytest.StashKey[RampartSession]()
 _session_start_key = pytest.StashKey[float]()
+_collector_key = pytest.StashKey[ResultCollector]()
 
 # Module-level constants are an acceptable exception in a hook-based
 # plugin module where there is no natural owning class.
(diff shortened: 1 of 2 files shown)


rampart/pytest_plugin/plugin.py
```diff
@@ -413,6 +415,81 @@ def _absorb_results(
...
+            summary="Trial completed without recording a Result.",
+        )
+    else:
+        summary = str(call.excinfo.value) or "Trial assertion did not pass."
+        result = Result(
+            safe=False,
+            status=SafetyStatus.UNDETERMINED,
+            summary=summary,
+        )
+    collector.record(result=result)
```
spencrr, line 445: this part makes me a bit uneasy - would also have to check if we are forwarding all the other metadata fields like category, etc.

rampart/pytest_plugin/plugin.py
```diff
@@ -413,6 +415,81 @@ def _absorb_results(
...
+        not report.failed
+        or call.excinfo is None
+        or not call.excinfo.errisinstance(AssertionError)
+    ):
+        return report
+
+    _record_missing_trial_result(collector=collector, call=call)
+    report.outcome = "passed"
+    report.longrepr = None
+    return report
```
spencrr, line 490: I also am not sure about how we want to think about the population number total reported by pytest: if I have a suite of 10 tests, with differing number of trials, we would potentially be weighting those tests as it currently stands - perhaps more thought about hooking into execution loop to run our own loop and then report to pytest just 1 result (akin to [pytest-loop](https://github.com/anogowski/pytest-loop/blob/master/src/pytest_loop/__init__.py#L103))

behnam-o:
this was opened to allow overriding pytest exit code when implementing trials + threshold in their current state. closing in favor of the work that will introduce an alternative way to achieve that

PR #120, opened 2026-07-15
[FEAT]: Add structured prompt template validation

spencrr:
## Description

Adds a structured `PromptTemplate` model with validated YAML loading and rendering.

- Replaces bare `jinja2.Template` loading with `PromptTemplate.from_yaml()`.
- Validates YAML through a strict private Pydantic schema.
- Rejects missing and unexpected render arguments.
- Verifies declared parameters against Jinja AST variables.
- Migrates `LLMDriver` and `LLMJudge` to the structured template API.
- Adds coverage for rendering, schema failures, parameter drift, block scalars, malformed YAML, and non-mapping YAML.

## Breaking changes

`load_prompt_template()` has been removed in favor of `PromptTemplate.from_yaml()`.

The loader now returns a `PromptTemplate` rather than a `jinja2.Template`; callers using Jinja-specific APIs must migrate to the structured API.

## Checklist

- [x] Tests added or updated for structured loading, rendering validation, schema failures, and raw YAML
- [x] Documentation updated in the prompt-template module docstring

### DIFF
--- a/rampart/common/templates.py
+++ b/rampart/common/templates.py
@@ -91,7 +91,7 @@ class TemplateParameterError(ValueError):
         super().__init__(msg)
 
 
-@dataclass(frozen=True, slots=True)
+@dataclass(frozen=True, kw_only=True, slots=True)
 class PromptTemplate:
     """Compiled prompt template with metadata and an explicit render contract."""
 
--- a/rampart/common/templates.py
+++ b/rampart/common/templates.py
@@ -19,57 +19,30 @@ render call before evaluating Jinja markup.
 
 from __future__ import annotations
 
+from collections.abc import Hashable
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Annotated, TypeAlias, TypeVar
 
 import yaml
 from jinja2 import Environment, StrictUndefined, Template, meta, select_autoescape
-from pydantic import BaseModel, ConfigDict, ValidationError, field_validator
+from pydantic import AfterValidator, BaseModel, ConfigDict, ValidationError
+
+__all__ = [
+    "PromptTemplate",
+    "PromptTemplateSchemaError",
+    "TemplateParameterError",
+]
 
 if TYPE_CHECKING:
     from collections.abc import Collection
     from pathlib import Path
+    from typing import Self
 
 
 _JINJA_ENVIRONMENT = Environment(
     autoescape=select_autoescape(default_for_string=False, default=False),
     undefined=StrictUndefined,
 )
-
-
-class _PromptTemplateYaml(BaseModel):
-    """Strict schema for the serialized YAML representation."""
-
-    model_config = ConfigDict(
-        extra="forbid",
-        frozen=True,
-        hide_input_in_errors=True,
-        strict=True,
-    )
-
-    name: str
-    parameters: list[str]
-    value: str
-    description: str | None = None
-
-    @field_validator("parameters")
-    @classmethod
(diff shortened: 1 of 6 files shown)


rampart/common/templates.py
```diff
@@ -4,43 +4,256 @@
...
+        if self.missing:
+            details.append(f"missing={self.missing!r}")
+        if self.unexpected:
+            details.append(f"unexpected={self.unexpected!r}")
+        joined_details = ", ".join(details)
+        msg = f"Prompt template {template_name!r} parameter mismatch: {joined_details}"
+        super().__init__(msg)
+
+
+@dataclass(frozen=True, slots=True)
```
nina-msft: nit: Our coding standard calls for `@dataclass(kw_only=True)` on classes with multiple fields. This one is `frozen=True, slots=True` but not `kw_only`, so the generated `__init__` is positional and exposes a leading-underscore positional arg `_template`. It works today only because `from_yaml` is the sole constructor and passes keywords.

Adding `kw_only=True` aligns with the standard and makes the private-field ordering moot. Minor.

rampart/common/templates.py
```diff
@@ -4,43 +4,256 @@
...
+    referenced_keys = meta.find_undeclared_variables(parsed)
+    declared_keys = set(definition.parameters)
+    if referenced_keys != declared_keys:
+        missing = tuple(sorted(referenced_keys - declared_keys))
+        unused = tuple(sorted(declared_keys - referenced_keys))
+        msg = (
+            f"Prompt template {definition.name!r} parameters do not match its "
+            f"Jinja variables: missing={missing!r}, unused={unused!r}"
+        )
+        raise ValueError(msg)
```
nina-msft: nit: This parameter-drift check raises a bare `ValueError`, but the schema check in `_load_yaml_definition` raises the dedicated `PromptTemplateSchemaError`. Both are load-time template-authoring failures, so a caller that wants to catch "this template file is broken" has to catch the typed error *and* a plain `ValueError`.

Consider raising a typed error here too — either reuse `PromptTemplateSchemaError` or add a sibling (e.g. `PromptTemplateContractError`) — so all `from_yaml` failures share one catchable surface. Not blocking; fine as a follow-up.

nina-msft (dismissed):
Great improvement!

rampart/common/templates.py
spencrr: Done.

rampart/common/templates.py
spencrr: Replaced the schema-specific error with `PromptTemplateDefinitionError`, covering invalid UTF-8, YAML, schema, Jinja, and parameter-contract content while preserving underlying causes.

rampart/common/templates.py
```diff
@@ -4,43 +4,270 @@
...
+        msg = (
+            f"Prompt template {definition.name!r} parameters do not match its "
+            f"Jinja variables: missing={missing!r}, unused={unused!r}"
+        )
+        raise PromptTemplateDefinitionError(path=path, details=msg)
+
+    try:
+        return _JINJA_ENVIRONMENT.from_string(parsed)
+    except TemplateError as exc:
+        raise PromptTemplateDefinitionError(path=path, details=str(exc)) from exc
```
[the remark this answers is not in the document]
spencrr: @copilot

```
    def from_string(
        self,
        source: t.Union[str, nodes.Template],
        globals: t.Optional[t.MutableMapping[str, t.Any]] = None,
        template_class: t.Optional[t.Type["Template"]] = None,
    ) -> "Template":
```

nina-msft (approved):

PR #121, opened 2026-07-20
[FEAT]: Add execution trial populations and threshold verdicts

behnam-o:
- Add `execute_trials_async()` and public `PopulationResult`.
- Evaluate repeated executions using configurable safety thresholds.
- Preserve population metadata across pytest, xdist, and JSON reporting.
- Update documentation and unit tests.

## Proposed stack order

1. #121 (this PR)
2. #123

Merge this PR first. GitHub cannot link these PRs as an official stack because both heads are in a fork.


rampart/core/result.py
```diff
@@ -158,6 +158,87 @@ def __repr__(self) -> str:
...
+
+    @property
+    def executed_count(self) -> int:
+        """Number of executed trials."""
+        return len(self.results)
+
+    @property
+    def total_count(self) -> int:
+        """Number of trials in the population."""
+        return self.executed_count
```
bashirpartovi: I think this is dead code, since it only aliases `executed_count` and adds no value. Plus it doesn't have any consumers

rampart/pytest_plugin/_session.py
```diff
@@ -238,13 +238,13 @@ def record_trial_group(
...
+                - ERROR results make the group fail.
+                - threshold is the minimum pass rate (SAFE / executed).
           e.g. 0.8 means at least 80% of runs must be SAFE.
-        - ERROR results count against the pass rate (they're not SAFE).
         - Clones with zero results (skipped or crashed before producing
-          a Result) are tracked as ``no_result`` and count against
-          the pass rate.
+                    a Result) are tracked as ``no_result`` and excluded from
+                    the pass-rate denominator.
+                - UNSAFE and UNDETERMINED results count against the pass rate.
```
bashirpartovi: This docstring is not styled properly, it's garbled up

rampart/pytest_plugin/_session.py
```diff
@@ -276,8 +276,13 @@ def record_trial_group(
...
 
-        pass_rate = safe_count / total if total > 0 else 0.0
-        passed = unsafe_count == 0 and pass_rate >= threshold
+        executed_count = total - no_result_count
+        pass_rate = safe_count / executed_count if executed_count > 0 else 0.0
+        passed = (
+            error_count == 0
+            and executed_count > 0
+            and pass_rate >= threshold
+        )
```
bashirpartovi: I think you need to change this to pass `ruff`
```python
passed = error_count == 0 and executed_count > 0 and pass_rate >= threshold
```

rampart/core/execution.py
```diff
@@ -273,6 +273,54 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
...
+            threshold (float): Required safe-result rate from 0.0 to 1.0.
+
+        Returns:
+            PopulationResult: Aggregate verdict and individual trial results.
+
+        Raises:
+            TypeError: If n is not a non-boolean integer.
+            ValueError: If n is less than 1 or threshold is outside
+                [0.0, 1.0].
+        """
```
bashirpartovi: Can we add one line to this docstring that says:
> Trials are only statistically meaningful when the adapter is stateless across sessions; a stateful adapter (e.g. memory-backed) makes `pass_rate` an unreliable estimate.

rampart/core/result.py
```diff
@@ -158,6 +158,87 @@ def __repr__(self) -> str:
...
+        Raises:
+            ValueError: If threshold is outside [0.0, 1.0].
+        """
+        if not 0.0 <= self.threshold <= 1.0:
+            msg = "threshold must be between 0.0 and 1.0"
+            raise ValueError(msg)
+    @property
+    def safe_count(self) -> int:
+        """Number of safe trials."""
+        return sum(result.status is SafetyStatus.SAFE for result in self.results)
```
bashirpartovi: use `result.safe` instead of `SafetyStatus`. 
```python
return sum(1 for result in self.results if result.safe)
```
This way you won't diverge from how safety is calculated for `Result`.

bashirpartovi: We probably need a guard in `Result` to protect against consumer checking `SafetyStatus` manually, e.g. 
```python
    ...
    metadata: dict[str, Any] = field(default_factory=dict[str, Any])

    def __post_init__(self) -> None:
        if self.safe != (self.status is SafetyStatus.SAFE):
            msg = (
                f"Result.safe={self.safe} contradicts "
                f"status={self.status.value}"
            )
            raise ValueError(msg)

    @property
    def eval_results(self) -> list[EvalResult]:
    ...
```

rampart/core/execution.py
```diff
@@ -273,6 +273,54 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
         )
         return result
 
+    async def execute_trials_async(
```
bashirpartovi: I think there is a problem we need to address because only the child `Result`s are auto‑recorded, the `PopulationResult` and its threshold and grouping is discarded when the test body returns. We may want to capture that in the report.

bashirpartovi: Maybe we could do something like this:
```python
import uuid  

    ...
    population_id = uuid.uuid4().hex
    results: list[Result] = []
    for index in range(n):
        result = await self.execute_async(adapter=adapter)
        result.metadata.setdefault(
            "_rampart_population",
            {
                "id": population_id,
                "threshold": threshold,
                "n": n,
                "index": index,
            },
        )
        results.append(result)

    return PopulationResult(results=results, threshold=threshold)
```
Although we need to think about this a bit more. uuid avoids population conflict but varies run to run..

rampart/core/result.py
```diff
@@ -158,6 +158,87 @@ def __repr__(self) -> str:
         )
 
 
+@dataclass(kw_only=True)
+class PopulationResult:
```
bashirpartovi, line 199: Can we add a `__repr__` to `PopulationResult` like `Result`? A compact one

rampart/core/execution.py
```diff
@@ -273,6 +273,54 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
...
+        Returns:
+            PopulationResult: Aggregate verdict and individual trial results.
+
+        Raises:
+            TypeError: If n is not a non-boolean integer.
+            ValueError: If n is less than 1 or threshold is outside
+                [0.0, 1.0].
+        """
+        if not isinstance(n, int) or isinstance(n, bool):
+            msg = "n must be an integer"
```
bashirpartovi: `True` is technically an integer. Maybe we should be consistent with the docstring? 
```python
msg = " n must be a non-boolean integer"
```

rampart/core/result.py
```diff
@@ -158,6 +158,87 @@ def __repr__(self) -> str:
...
+    def safe(self) -> bool:
+        """Whether the population met its safety threshold."""
+        return self.status is SafetyStatus.SAFE
+
+    @property
+    def summary(self) -> str:
+        """Concise population verdict summary."""
+        return (
+            f"{self.safe_count}/{self.executed_count} trials safe "
+            f"({self.pass_rate:.0%} pass rate, threshold: {self.threshold:.0%}); "
```
bashirpartovi: Maybe we should use `.1%`? because for things like `0.855` this will translate to `86%` instead of `85.5%`

tests/unit/core/test_result.py
```diff
@@ -135,6 +145,74 @@ def test_harm_category_accepts_plain_string(self) -> None:
         assert r.harm_category == "custom_product_risk"
 
 
+class TestPopulationResult:
```
bashirpartovi, line 184: Can we add a test for when status is all error or all undetermined?

rampart/core/result.py
spencrr: Can we just change to:

```python
class Result:
...
  @property
  def safe(self):
    return self.status is SafetyStatus.SAFE
```

Can also keep `Result.__bool__` as `return self.safe` too and avoids incorrect init

bashirpartovi: @spencrr that would have been a great choice but unfortunately not with the current repo setup. e.g. 
https://github.com/microsoft/RAMPART/blob/83b1ad593fa8a62b1e3f4aae5a301fc41fd38679/rampart/attacks/_xpia.py#L203
`safe` is also part of the dataclass and can be set at runtime. For this PR, it should go into post-init till we can discuss what to do with the dataclass field. If we remove it, your solution would be ideal to implement.

bashirpartovi: Although, I think we should drop it from the dataclass in a separate PR and clean the call sites.

behnam-o: This is a good observation, and I like to keep it consistent with PopulationResult (i.e. safe is a derived property instead of being settable, and always checked)

but it seems like out of scope for this PR, so let's do it on its own here: https://github.com/microsoft/RAMPART/pull/124

rampart/core/execution.py
behnam-o: @bashirpartovi good callout. 

I liked your metadata suggestion and implemented it with a minor tweak: execute_async() now accepts additional result metadata and attaches it before ON_POST_EXECUTE fires. This ensures every event handler, including the built-in result collector, observes the final annotated Result, and the metadata flows through to reporting sinks by default.

let me know what you think

behnam-o: also about the uuid - I think it's actuallya good think to use new ones per population, because it only groups results from one execute_trials_async() invocation; stable pytest context such as _pytest_nodeid and _rampart_result_index is persisted separately for cross-run identification and ordering.

rampart/core/result.py
```diff
@@ -168,6 +168,98 @@ def __repr__(self) -> str:
...
+    results: list[Result]
+    threshold: float
+
+    def __post_init__(self) -> None:
+        """Validate population configuration.
+
+        Raises:
+            ValueError: If threshold is outside [0.0, 1.0].
+        """
+        if not 0.0 <= self.threshold <= 1.0:
```
spencrr, line 224: just verifying this is equivalent to ```not(0.0 <= self.threshold and self.threshold <= 1.0)```

rampart/core/execution.py
```diff
@@ -272,6 +293,69 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
...
+                additional_result_metadata={
+                    "_rampart_population": {
+                        "id": population_id,
+                        "index": index,
+                        "size": n,
+                        "threshold": threshold,
+                    },
+                },
+            )
+            results.append(result)
```
spencrr: do we wait to await all here?

rampart/pytest_plugin/_session.py
```diff
@@ -238,13 +238,13 @@ def record_trial_group(
...
-        - Clones with zero results (skipped or crashed before producing
-          a Result) are tracked as ``no_result`` and count against
-          the pass rate.
+            - ERROR results make the group fail.
+            - Threshold is the minimum pass rate (SAFE / executed); e.g.,
+                0.8 means at least 80% of runs must be SAFE.
+            - Clones with zero results (skipped or crashed before producing
+                a Result) are tracked as ``no_result`` and excluded from the
+                pass-rate denominator.
+            - UNSAFE and UNDETERMINED results count against the pass rate.
```
spencrr: just checking on the semantics here: so error-ed trials now do not count toward a failure -> does this mean if we have errors, denominator is decreased and test passes?

rampart/core/execution.py
spencrr: or as an opt-in option?

docs/usage/results-and-reporting.md
```diff
@@ -129,22 +129,22 @@ For CI gating, capture a curated set of facts in `result.metadata` — both scen
...
+        "mitigation_ref": "SEC-1234",
+        "ci_run_url": "https://ci.example.com/runs/94821",
+    },
+)
 
 assert result, result.summary
 ```
 
-These keys live on the `Result`, so any sink _can_ persist them. With `JsonFileReportSink`, for example, they appear on each result's `metadata` object (grouped under `by_harm_category` in the output). A custom sink only records them if its `emit_async` reads `result.metadata`.
+Additional metadata is attached before `ON_POST_EXECUTE`, so event handlers and sinks see the same result state. It is strictly additive: reusing a key already produced by the execution raises `ValueError`. Keys beginning with `_rampart_` are conventionally used by the framework and should be avoided by callers. With `JsonFileReportSink`, these keys appear on each result's `metadata` object (grouped under `by_harm_category` in the output). A custom sink only records them if its `emit_async` reads `result.metadata`.
```
spencrr: likely we can move the rampart internal metadata later so that we don't have to worry about these namespace collisions, but this is good for now

spencrr (dismissed):
LGTM - let's also check with @bashirpartovi ! One question about the semantics of failed tests in pass/fail calculation but everything else makes sense to me and test coverage is good. Thanks @behnam-o !

rampart/core/execution.py
```diff
@@ -264,6 +281,10 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
 
         elapsed = time.monotonic() - start
         result.duration_seconds = elapsed
+        self._add_result_metadata(
+            result=result,
+            additional_result_metadata=additional_result_metadata,
+        )
```
bashirpartovi: This can completely terminate entire population data. `_add_result_metadata` raises `ValueError` on duplicate key, but this is outside the `try/except` clause so any error raised bypasses `ON_POST_EXECUTE` and `ON_ERROR` events, therefore, the result is never collected and no error result is produced. This just crashes the entire run. And since this is used in `execute_trials_async`, the blast radius is the whole population, all prior trials are discarded. `test_rejects_duplicate_additional_metadata_async` already pins this behavior, it asserts only `ON_PRE_EXECUTE` fired.

This isn't only programmer error either, it's reachable from adapter data. `_collect_response_metadata` promotes `t.response.metadata` straight to the top level of `Result.metadata` whenever exactly one turn carries metadata, and its own docstring says adapters attach things like `conversation_id` and `session_id`. `results-and-reporting.md` now tells users to pass keys like `scenario_id` and `session_id` through `additional_result_metadata`, so a collision is easy to hit by accident rather than something the caller can always predict.

Can we move the `_add_result_metadata` call inside the `try/except` so a collision degrades to an `ERROR` result like every other failure? At minimum the framework's own `_rampart_population` stamping shouldn't be able to abort a run.

rampart/core/execution.py
```diff
@@ -272,6 +293,69 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
...
+        """Execute a population of independent trials.
+
+        Each trial uses the normal ``execute_async`` lifecycle, including
+        event dispatch and result collection. The returned aggregate provides
+        the single logical verdict that callers should assert. Execution
+        strategies are responsible for creating a fresh agent session during
+        each call to ``execute_async``.
+
+        Note: Trials are only statistically meaningful when the adapter is stateless
+        across sessions. a stateful adapter (e.g. memory-backed) makes pass_rate an
```
bashirpartovi: typo: 
> .. sessions. a stateful ..

"a" should be capitalized

bashirpartovi (commented):
Just to step back a little bit, I think `additional_result_metadata` on `execute_async` is doing two jobs that want opposite things. `_rampart_population` is framework-internal provenance that should never be able to fail, and the scenario keys the docs now recommend (`scenario_id`, `threat_class`, and so on) are caller data where failing loudly on a typo is fine. Both go into the same flat dict with strict duplicate rejection, and that's what produces the failure mode.

It's also a second convention for something we already do. `absorb()` injects framework keys into `result.metadata` after the fact, last write wins, and it never raises:

```python
result.metadata = {
    **original_result.metadata,
    "_pytest_test_name": test_name,
    "_pytest_nodeid": node.nodeid,
    "_rampart_result_index": result_index,
}
```

So we already have a pattern for framework-internal metadata and this adds a stricter one next to it.

I'm also not sure the parameter buys much over writing to metadata after the call. The reason for a parameter is that handlers see it at ON_POST_EXECUTE, but `ResultCollectionHandler` holds a reference, the collector drains at teardown, and xdist serializes at absorb, so a post-hoc write is already visible everywhere that matters. The only thing that misses it is a custom handler reading `metadata` inside `on_event`, and population identity is a property of the aggregate rather than of one execution anyway.

The thing that convinced me the parameter is on the wrong object is that callers can't use it on the trials path. `execute_trials_async(adapter, n, threshold)` has no `additional_result_metadata`, and the only thing that fills that argument there is the framework's own `_rampart_population` dict. So the receipt pattern we just documented works for a single run and quietly stops being available once you switch to trials:

```python
# works
result = await Attacks.xpia(...).execute_async(
    adapter=my_agent,
    additional_result_metadata={"scenario_id": "xpia-login-001"},
)

# no way to say the same thing here
population = await Attacks.xpia(...).execute_trials_async(
    adapter=my_agent, n=10, threshold=0.8,
)
```

Supporting it would mean adding the parameter to `execute_trials_async` as well and merging the caller's dict with `_rampart_population` before passing it down, which is more surface and two sources feeding one argument.

Scenario metadata is constant across every trial anyway, so it seems like it belongs where the scenario is declared rather than on each call:

```python
attack = Attacks.xpia(
    inject=...,
    trigger="Summarize Q3",
    evaluator=...,
    metadata={
        "scenario_id": "xpia-login-001",
        "threat_class": "credential_exfiltration",
        "mitigation_ref": "SEC-1234",
    },
)

result = await attack.execute_async(adapter=my_agent)
population = await attack.execute_trials_async(adapter=my_agent, n=10, threshold=0.8)
```

The execution copies it onto each `Result` it produces, so it works the same for one run and for a population and neither method grows a parameter. Run level stuff like `ci_run_url` already has `report.metadata` and probably shouldn't be duplicated onto every result.

That leaves the population provenance, and I'd rather see it as a typed field than a reserved dict key. We already do this with `injections: list[InjectionRecord]`, which is XPIA-specific provenance carried as a real field and documented as empty for non-XPIA tests:

```python
@dataclass(kw_only=True)
class PopulationRef:
    """Identifies the trial population a Result belongs to."""

    id: str
    index: int
    size: int
    threshold: float


# on Result, next to injections
population: PopulationRef | None = None
```

and the loop sets it directly, the same way `execute_async` already sets `duration_seconds`:

```python
population_id = uuid.uuid4().hex
results: list[Result] = []
for index in range(n):
    result = await self.execute_async(adapter=adapter)
    result.population = PopulationRef(
        id=population_id, index=index, size=n, threshold=threshold,
    )
    results.append(result)
```

Collisions become impossible, `ty` checks it, and WS0-05 gets to serialize a known field instead of a reserved key. If the typed field is too much for this PR, the smaller version is to drop the parameter and stamp into metadata in that same spot:

```python
result.metadata["_rampart_population"] = {
    "id": population_id, "index": index, "size": n, "threshold": threshold,
}
```

which matches what `absorb()` already does and can't raise.

rampart/core/execution.py
behnam-o: Good points. I replaced `additional_result_metadata` with a typed `PopulationRef` on `Result`, avoiding collisions and working consistently across serialization. Population provenance is attached through a shared lifecycle helper before `ON_POST_EXECUTE`, so handlers still see it. Generic scenario metadata remains separate and can be addressed in a follow-up API discussion.

rampart/core/result.py
behnam-o: yup - [ref](https://docs.python.org/3/reference/expressions.html#:~:text=Comparisons%20can%20be%20chained%20arbitrarily%2C%20e.g.%2C%20x%20%3C%20y%20%3C%3D%20z%20is%20equivalent%20to%20x%20%3C%20y%20and%20y%20%3C%3D%20z)

>Comparisons can be chained arbitrarily, e.g., x < y <= z is equivalent to x < y and y <= z

spencrr: got it - wasn't sure if the not applied to the whole chain

spencrr: coming from C, this looks like not(0) = 1 < threshold < 1

rampart/pytest_plugin/_session.py
behnam-o: this made me realize we probably don't wanna change these semantics at all ... I reverted them back to what they were before. 

I think the whole idea of trial groups and this aggregate stats on them will go away after this set of PRs (this and #123 ) complete.

rampart/core/execution.py
behnam-o: good idea. added support to run these in parallel, but default is sequential.

rampart/core/execution.py
```diff
@@ -281,20 +286,30 @@ async def execute_trials_async(
...
+                        id=population_id,
+                        index=index,
+                        size=n,
+                        threshold=threshold,
+                    ),
+                    semaphore=semaphore,
+                )
+                for index in range(n)
+            ),
+        )
```
spencrr: ai suggests a different approach here -- what do you think?

```python
population_id = uuid.uuid4().hex
semaphore = asyncio.Semaphore(max_concurrency)
async with asyncio.TaskGroup() as tg:
    tasks = [
        tg.create_task(
            self._execute_trial_async(
                adapter=adapter,
                population=PopulationRef(
                    id=population_id,
                    index=index,
                    size=n,
                    threshold=threshold,
                ),
                semaphore=semaphore,
            )
        )
        for index in range(n)
    ]
results = [task.result() for task in tasks]
```

<details>
<summary>AI Reasoning</summary>

## Reasoning

Four things to check against the asyncio model: (a) does the bound actually bound, (b) what happens on failure, (c) is ordering preserved, (d) is the API shape idiomatic.

**a) The bound works, the task count doesn't.** `Semaphore(max_concurrency)` acquired inside each coroutine correctly caps *in-flight* trials. But `asyncio.gather(*(...))` eagerly wraps all `n` coroutines into Tasks immediately — so `n=10_000, max_concurrency=4` allocates 10k Task objects up front, 9,996 of which sit parked on the semaphore. Bounded concurrency, unbounded memory. Fine for `n` in the tens; a footgun if `n` is ever user-supplied and large.

**b) Failure handling is the real bug.** `gather()` without `return_exceptions=True` re-raises the first exception *as soon as it fires*, but **does not cancel the sibling tasks** — they keep running detached from any awaiter. In a harness that spawns agent sessions, that means orphaned sessions, leaked subprocess/HTTP handles, and events dispatched after `execute_trials_async` has already raised. This is documented behavior: "if `gather()` is cancelled, awaitables are cancelled" — but the reverse (a child raising) does not propagate cancellation downward.

This is exactly what `asyncio.TaskGroup` (3.11+) was added to fix: it cancels remaining children on first failure and won't exit until all are done.

**c) Ordering is fine.** `gather` returns results in argument order regardless of completion order, so `results[i].population.index == i` still holds. Good. Note the *event stream* is no longer ordered though — if any consumer of `_fire` assumes trials arrive sequentially, that assumption silently breaks at `max_concurrency > 1`. Worth an explicit note in the docstring.

**d) API shape.** `max_concurrency: int = 1` with `>= 1` validation is defensible (backward-compatible default), but there's no way to say "unbounded." Convention is `None` = unlimited, or `<= 0` = unlimited.

Confidence: **High** on (b) — this is documented `gather` semantics. **Medium** on (a) being worth fixing; depends entirely on your expected `n`.

## Deliverable

Minimum fix — swap `gather` for a TaskGroup so failures cancel siblings:

```python name=rampart/core/execution.py
population_id = uuid.uuid4().hex
semaphore = asyncio.Semaphore(max_concurrency)
async with asyncio.TaskGroup() as tg:
    tasks = [
        tg.create_task(
            self._execute_trial_async(
                adapter=adapter,
                population=PopulationRef(
                    id=population_id,
                    index=index,
                    size=n,
                    threshold=threshold,
                ),
                semaphore=semaphore,
            )
        )
        for index in range(n)
    ]
results = [task.result() for task in tasks]
```

If you're on 3.9/3.10, the portable equivalent is `gather(..., return_exceptions=True)` plus an explicit cancel-and-drain of the remaining tasks before re-raising — don't leave bare `gather`.

If `n` can be large, prefer a fixed worker pool over `n` parked tasks — this drops the semaphore entirely and caps task count at `max_concurrency`:

```python name=rampart/core/execution.py
results: list[Result | None] = [None] * n
indices = iter(range(n))

async def worker() -> None:
    for index in indices:  # iterator is atomic under the single-threaded loop
        results[index] = await self._execute_once_async(
            adapter=adapter,
            population=PopulationRef(
                id=population_id, index=index, size=n, threshold=threshold
            ),
        )

async with asyncio.TaskGroup() as tg:
    for _ in range(min(max_concurrency, n)):
        tg.create_task(worker())
```

Two smaller items:

- **Unbounded escape hatch.** Accept `max_concurrency: int | None = 1`, treat `None` as `n`. Keeps the `>= 1` validation for ints and gives callers a way to say "all at once" without passing a magic number.
- **Docstring warning.** The existing stateless-adapter note gets sharper teeth under concurrency — a stateful adapter now has *interleaved* corruption, not just cross-trial leakage. Also flag that event dispatch order is non-deterministic when `max_concurrency > 1`, so downstream consumers should key off `population.index` rather than arrival order.

One thing I can't verify from the diff: whether `_fire` / the event sink is safe under concurrent invocation. If it appends to a shared list or writes to a stream, it's fine on a single-threaded loop as long as there's no `await` mid-mutation — worth a look before merging.
</details>

behnam-o: Thanks for the suggestion! looks like there's some benefits to task groups, updated to use those instead.

rampart/core/execution.py
```diff
@@ -230,6 +236,111 @@ async def execute_async(self, *, adapter: AgentAdapter) -> Result:
...
+                    self._execute_trial_async(
+                        adapter=adapter,
+                        population=PopulationRef(
+                            id=population_id,
+                            index=index,
+                            size=n,
+                            threshold=threshold,
+                        ),
+                        semaphore=semaphore,
+                    ),
```
bashirpartovi: This doesn't really run independent trials because every trial uses the same `BaseExecution` object. A fresh agent session gets created each time, but everything else on the execution object, like the driver, evaluator, handlers, and injection handles, gets reused. That means state from one trial can affect another.

For example, consider this case:

1. A factory creates one `SingleTurnExecution` or `XPIAExecution`.
2. That execution stores one driver in `self._driver`.
3. `execute_trials_async()` repeatedly calls `_execute_async()` on that same execution.
4. Each call creates a fresh agent session, but the driver is not recreated.
5. `LLMDriver` has one `_conversation_id`, and its messages are kept in PyRIT's `CentralMemory`.
6. Trial 2 starts with a fresh agent-side `history=[]`, but the driver-side conversation still has trial 1.
7. `LLMDriver._assert_conversations_consistent()` raises `DriverError`.
8. `BaseExecution` turns that exception into an `ERROR` result, so a population that should contain independent trials becomes `[SAFE, ERROR]`.

The same problem can happen with anything else that keeps state between runs.

I think each trial needs at least:

- A fresh agent session
- A fresh driver with a new conversation ID
- Fresh injection handles, unless we decide to activate them once for the whole population
- A fresh evaluator if it keeps any state

Things that don't keep state can be shared. Anything that does keep state should be recreated for each trial unless we know it was designed to be reused.

One possible way to handle this would be to keep the population loop in `BaseExecution`, but let each execution type create a fresh execution for every trial:

```python
class BaseExecution:
    async def execute_trials_async(...):
        executions = [
            self._create_trial_execution()
            for _ in range(n)
        ]
        ...
```

Then `SingleTurnExecution` could create a new driver through a factory instead of reusing `self._driver`:

```python
class SingleTurnExecution:
    def _create_trial_execution(self) -> BaseExecution:
        return SingleTurnExecution(
            driver=self._driver_factory(),
            ....
        )
```

If callers can provide stateful drivers, the public API may need to accept a factory instead of an existing driver object:

```python
Probes.behavior(
    driver_factory=lambda: LLMDriver(...),
    evaluator=evaluator,
)
```

I don't think we can safely rebuild a driver with `deepcopy()`. Drivers can hold model clients, locks, normalizers, and external memory state that should not be copied.

For XPIA, I think we need to pick one of these options:

- Create and activate fresh injection handles for every trial
- Activate one injection around the whole population, then create fresh sessions and drivers for each trial

Either option could work, but repeatedly entering the same handle is not safe.

I would also remove `max_concurrency` for now. It makes this problem worse because multiple trials start using the same stateful objects at the same time.

Could we also add tests with an `LLMDriver` and a handle that cannot be entered twice? The current isolation test proves that sessions are different, but it uses `StaticDriver`, so it doesn't prove that the full trial run is independent.

rampart/core/result.py
```diff
@@ -168,6 +187,98 @@ def __repr__(self) -> str:
...
+    def pass_rate(self) -> float:
+        """Safe-result rate across executed trials."""
+        if self.executed_count == 0:
+            return 0.0
+        return self.safe_count / self.executed_count
+
+    @property
+    def status(self) -> SafetyStatus:
+        """Population status resolved using error and threshold policy."""
+        if any(result.status is SafetyStatus.ERROR for result in self.results):
```
bashirpartovi, line 248: This makes sense to me as long as the follow-up PR removes the clone aggregation path before these changes are released. Right now `PopulationResult` and `record_trial_group()` give different answers for the same population, but hopefully that difference goes away once the marker becomes declaration-only and all trial verdicts go through `PopulationResult`.

rampart/pytest_plugin/_xdist.py
```diff
@@ -1161,6 +1173,16 @@ def _deserialize_result(*, data: object) -> Result:
                 raw_injections if isinstance(raw_injections, list) else [],
             )
         ],
+        population=(
+            PopulationRef(
+                id=str(raw_population.get("id", "")),
+                index=int(raw_population.get("index", 0)),
+                size=int(raw_population.get("size", 0)),
+                threshold=float(raw_population.get("threshold", 0.0)),
+            )
```
bashirpartovi: There is a small xdist error-handling gap here. These `int()` and `float()` calls can raise `ValueError` or `TypeError` for malformed population data, but the controller only catches `WorkerOutputError`. For example, `index="bad"` raises a raw `ValueError` instead of marking the run incomplete like the rest of the xdist validation does.

I don’t think this is a blocker because `execute_trials_async()` creates valid population values and agent output cannot control these fields. It can still happen if someone manually constructs a `PopulationRef`, though, since that is a public type and currently does not validate its values.

Could we parse this through a small helper that checks the ID, index, size, and threshold and turns parsing failures into `WorkerOutputError`? I think `_deserialize_result()` should either return a valid `PopulationRef` or fail through the same error path as the rest of the worker payload.

bashirpartovi (commented):
Thanks @behnam-o for working through all the feedback but I think there is a concurrency issue with the trial run that should get fixed before merging this. Let me know if it doesn't make sense.

rampart/core/execution.py
behnam-o: Thanks for pointing this out. I had a design dilleman how to address this, tried a bunch of patterns, and committed option 2 below - let me know if you think it's a good choice. 

## Trial Isolation Design

Trials must be independent, but an execution can contain stateful drivers, evaluators, handlers, and injection handles. Reusing these objects may make trials co-dependent.

### Option 1: Factory for Every Stateful Dependency

```python
execution = Probes.behavior(
    driver_factory=lambda: LLMDriver(...),
    evaluator_factory=lambda: MyEvaluator(),
    event_handler_factories=[
        lambda: MyHandler(),
    ],
)

population = await execution.execute_trials_async(
    adapter=adapter,
    n=10,
    threshold=0.8,
)
```

The execution constructs fresh dependencies internally:

```python
class SingleTurnExecution(BaseExecution):
    async def _execute_async(self, *, adapter):
        driver = self._driver_factory()
        evaluator = self._evaluator_factory()
        ...
```

**Pros**

- Preserves `execution.execute_trials_async(...)`.
- Explicitly creates fresh dependencies.

**Cons**

- Requires identifying every stateful constructor argument.
- APIs accumulate `driver_factory`, `evaluator_factory`, `handle_factory`, etc.
- A newly added stateful dependency can easily be overlooked.
- Lifecycle policy becomes duplicated across execution implementations.

### Option 2: Factory for the Complete Execution

```python
def create_execution() -> BaseExecution:
    return Probes.behavior(
        driver=LLMDriver(...),
        evaluator=MyEvaluator(),
        event_handlers=[MyHandler()],
    )

population = await execute_trials_async(
    execution_factory=create_execution,
    adapter=adapter,
    n=10,
    threshold=0.8,
)
```

The framework invokes the factory for every trial:

```python
async def run_trial(index: int) -> Result:
    execution = execution_factory()

    return await execution.execute_async(
        adapter=adapter,
    )
```

XPIA follows the same pattern:

```python
def create_execution() -> BaseExecution:
    return Attacks.xpia(
        inject=surface.inject(payload=payload),
        trigger=LLMDriver(...),
        evaluator=MyEvaluator(),
    )
```

**Pros**

- Creates the complete execution graph per trial.
- Covers future constructor parameters automatically.
- Avoids unsafe copying.
- Avoids dependency-specific factory proliferation.
- Makes the independence requirement visible at the call site.
- Works consistently across execution implementations.

**Cons**

- Population tests use a standalone function instead of an instance method.
- Users can still accidentally close over shared mutable objects.

For example, this remains incorrect:

```python
shared_driver = LLMDriver(...)

def create_execution() -> BaseExecution:
    return Probes.behavior(
        driver=shared_driver,
        evaluator=MyEvaluator(),
    )
```

## Decision

We chose the **complete execution factory** because freshness applies to the whole execution graph, not a fixed list of known dependencies.

The resulting contract is:

> Each factory invocation must return a fresh execution with fresh trial-scoped dependencies.

Single-run usage remains unchanged:

```python
execution = Probes.behavior(
    driver=driver,
    evaluator=evaluator,
)

result = await execution.execute_async(adapter=adapter)
```

Only population execution requires an explicit construction strategy.

rampart/core/result.py
behnam-o: I assume you mean `TrialGroupResult.passed`, since `record_trial_group()` itself returns `None`. I aligned `TrialGroupResult.status` with `PopulationResult.status` and now derive `passed` from it, so both aggregation paths produce the same verdict.

but you're right, this will be completely removed in the follow up PR, just fixing here for consistency

rampart/pytest_plugin/_session.py
```diff
@@ -76,7 +104,27 @@ class TrialGroupResult:
     no_result: int
     threshold: float
     pass_rate: float
```
spencrr, line 106: nit: perhaps we derive total and pass_rate? can also rename to `TrialGroupResultStatistics` maybe since it's not the actual result object? no strong opinions on this though! (and sort of out of scope)

rampart/core/result.py
```diff
@@ -93,6 +93,23 @@ class InjectionRecord:
...
+        id: Unique identifier shared by every result in the population.
+        index: Zero-based position of the result within the population.
+        size: Number of results requested for the population.
+        threshold: Required safe-result rate for the population.
+    """
+
+    id: str
+    index: int
+    size: int
+    threshold: float
```
spencrr, line 110: Does this mean that each `Result` (that is part of a population) has duplicate size/threshold fields for each `Result` but really that's a shared field on the entire population?

rampart/reporting/json_file.py
```diff
@@ -120,6 +120,11 @@ def _serialize_result(self, result: Result) -> dict[str, Any]:
             "strategy": result.strategy,
             "observability_level": result.observability_level.value,
             "duration_seconds": result.duration_seconds,
+            "population": (
+                dataclasses.asdict(result.population)
+                if result.population is not None
+                else None
+            ),
```
spencrr, line 127: and would be printed out here

rampart/core/result.py
behnam-o: yeah technically we can just put a population_ref_id on the result, store/output/share the population_refs centrally and separately, and whoever wants to find correlated results, they can do the join- but that needs a different refactor in a few places. I think it doesn't hurt to have this duplicated content on all results.

rampart/pytest_plugin/_session.py
behnam-o: Right, it'll be removed in a follow-up PR that builds on this one, so I wouldn't worry about refactoring/optimizing this (I know I added the two properties, but I saw those as "missing bits")

rampart/reporting/json_file.py
behnam-o: replied. I don't mind only storing the id and index but that needs a change to reports to let users extract these. I'll defer to separate future PR

spencrr (approved):

PR #123, opened 2026-07-22
[FEAT]: Replace trial cloning with configurable trial populations

behnam-o:
## Summary

Replace pytest item cloning for `@pytest.mark.trial` with an explicit `trial_config` fixture. Tests control population execution while RAMPART supplies:

- The marker-declared population size and threshold
- A `--rampart-trials N` CLI override for population size
- Collection-time validation of trial declarations
- An error when a trial-marked test does not consume `trial_config`

This avoids clone-specific pytest and xdist behavior while preserving population-level result reporting.

## Changes

- Add the immutable `TrialConfig` public type
- Add the `trial_config` fixture
- Add the `--rampart-trials N` option
- Remove collection-time test cloning
- Validate trial markers during collection
- Update xdist tests for one-item, multiple-result populations
- Update trial documentation and examples

## Example

```python
from rampart import Probes, execute_trials_async

@pytest.mark.trial(n=10, threshold=0.8)
async def test_injection_resistance(adapter, trial_config):
    population = await execute_trials_async(
        execution_factory=lambda: Probes.behavior(...),
        adapter=adapter,
        n=trial_config.n,
        threshold=trial_config.threshold,
    )
    assert population, population.summary
```

This PR now includes the prerequisite execution-population work from #121 through the latest `main` merge.


rampart/pytest_plugin/plugin.py
spencrr, line 1: ahh much cleaner here! 🧼

rampart/pytest_plugin/plugin.py
```diff
@@ -240,162 +205,27 @@ def pytest_unconfigure(config: pytest.Config) -> None:
...
-                "clones may be split across workers. Aggregation remains "
-                "correct (controller merges all results), but using "
-                "--dist=loadgroup keeps trial clones co-located on one "
-                "worker for better locality.",
-                dist_mode,
-            )
+        resolve_trial_config(node=item, config=config)
+        if "trial_config" not in getattr(item, "fixturenames", ()):
+            msg = f"@pytest.mark.trial requires trial_config on {item.nodeid}"
+            raise pytest.UsageError(msg)
```
spencrr, line 228: I think this only enforces that the caller imports the fixture but not necessarily that they forward the args. Not that we necessarily need to strictly enforce this, but `execute_trials_async` could instead (1) take a `TrialConfig` (although this would require that every trial call site be fully configurable by CLI flag with required plumbing) or (2) refactor `execute_trials_async` as a fixture that depends on the `trial_config` fixture - what do you think?

behnam-o: Good point that consuming the fixture doesn’t guarantee its values are forwarded. I checked this against the WS0 design: WS0-01 intentionally keeps execute_trials_async(n, threshold) as a pytest-independent population API, while WS0-03 specifies a mismatch warning when the executed population differs from the marker declaration. Passing TrialConfig directly would change that core contract, and making execution a fixture would couple it to pytest. I think we should retain the scalar API and address incorrect forwarding through the planned mismatch warning instead.

spencrr (approved):

PR #132, opened 2026-07-27
[FIX]: Enforce PromptTemplate construction invariants

spencrr:
## Description

Fixes construction and equality invariants for `PromptTemplate` introduced in #120.

The dataclass-generated initializer exposed every field, including `_template`. A leading underscore is only a Python naming convention, so a caller could combine declared parameter metadata with an unrelated Jinja template and bypass `PromptTemplate.from_yaml()` validation:

```python
PromptTemplate(
    name="x",
    description=None,
    parameter_keys=("declared",),
    _template=Template("{{ actual }}"),
).render(declared="provided")
```

This renders an empty string: argument validation succeeds against `parameter_keys`, while the injected template references a different variable and was compiled outside RAMPART's `StrictUndefined` environment.

This PR:

- disables the generated dataclass initializer and explicitly rejects direct construction, including the otherwise-valid empty `PromptTemplate()` call left available by `init=False` alone;
- creates instances only from a schema-validated YAML definition;
- compiles the Jinja template inside the validated-definition factory so the declared parameters and compiled behavior cannot drift independently;
- uses `object.__new__()` and `object.__setattr__()` for the narrow initialization phase required by a frozen, slotted dataclass;
- disables generated structural equality, which previously treated templates with identical metadata but different bodies as equal because `_template` was excluded from comparison, while preserving identity-based hashing; and
- marks `PromptTemplate` as final for static type checkers and IDEs.

`PromptTemplate.from_yaml(path)` remains the supported construction API and retains its `Self` return type. Separate instances now use identity equality:

```python
first = PromptTemplate.from_yaml(path)
second = PromptTemplate.from_yaml(path)

assert first is not second
assert first != second
assert len({first, second}) == 2
```

## Breaking changes

Direct `PromptTemplate(...)` construction is now rejected. Use `PromptTemplate.from_yaml(path)` instead. Equality between separate instances is now identity-based rather than metadata-based. Existing RAMPART production call sites already construct templates through `from_yaml()`.

## Checklist

- [X] Tests added or updated for changes: direct construction, the former generated-constructor arguments, identity equality, and hashability
- [X] Documentation updated: factory-only construction and migration are documented above and in the API docstrings

### DIFF
--- a/rampart/common/templates.py
+++ b/rampart/common/templates.py
@@ -21,7 +21,7 @@ from __future__ import annotations
 
 from collections.abc import Hashable
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Annotated, TypeAlias, TypeVar
+from typing import TYPE_CHECKING, Annotated, TypeAlias, TypeVar, final
 
 import yaml
 from jinja2 import (
@@ -98,14 +98,30 @@ class TemplateParameterError(ValueError):
         super().__init__(msg)
 
 
-@dataclass(frozen=True, kw_only=True, slots=True)
+@final
+@dataclass(
+    frozen=True,
+    kw_only=True,
+    slots=True,
+    init=False,
+    eq=False,
+)
 class PromptTemplate:
     """Compiled prompt template with metadata and an explicit render contract."""
 
     name: str
     description: str | None
     parameter_keys: tuple[str, ...]
-    _template: Template = field(repr=False, compare=False)
+    _template: Template = field(repr=False)
+
+    def __init__(self) -> None:
+        """Reject construction that bypasses template validation.
+
+        Raises:
+            TypeError: Always. Use :meth:`from_yaml` to construct an instance.
+        """
+        msg = "Use PromptTemplate.from_yaml()"
+        raise TypeError(msg)
 
     @classmethod
     def from_yaml(cls, path: Path) -> Self:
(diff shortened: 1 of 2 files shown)


rampart/common/templates.py
```diff
@@ -127,12 +143,31 @@ def from_yaml(cls, path: Path) -> Self:
...
+    def _from_validated_definition(
+        cls,
+        *,
+        definition: _PromptTemplateYaml,
+        path: Path,
+    ) -> Self:
+        compiled = _compile_template(definition, path=path)
+
+        instance = object.__new__(cls)
+        object.__setattr__(instance, "name", definition.name)  # ruff:ignore[unnecessary-dunder-call]
```
nina-msft, line 158: Style nit:

This suppression form is inconsistent with the rest of the repo — everywhere else uses `# noqa: <CODE>` (e.g. `# noqa: PLC0415` in `converters/docx.py`, `# noqa: ANN401` in the evaluators). `# ruff:ignore[unnecessary-dunder-call]` keys off the rule *name* rather than the `PLC2801` code, and that `ignore[...]` bracket is actually ty's syntax (see `plugin.py:301`), not ruff's. It does suppress today, but it's easy to misread and a rule rename would silently stop suppressing. The sentinel refactor in the other comment removes all four of these entirely; if you keep this approach, please switch them to `# noqa: PLC2801` for consistency.

rampart/common/templates.py
```diff
@@ -127,12 +143,31 @@ def from_yaml(cls, path: Path) -> Self:
...
+            "description",
+            definition.description,
+        )
+        object.__setattr__(  # ruff:ignore[unnecessary-dunder-call]
+            instance,
+            "parameter_keys",
+            tuple(definition.parameters),
         )
+        object.__setattr__(instance, "_template", compiled)  # ruff:ignore[unnecessary-dunder-call]
+        return instance
```
nina-msft, line 170: Was considering your comment on whether `dataclass` is still necessary since we are limiting some of its core usage...what do you think about this approach?

The `object.__new__` + four `object.__setattr__` reconstruction here only exists to work around `init=False` on a frozen, slotted dataclass — the class opts into the dataclass machinery and then fights it. You can keep the dataclass (its generated `__init__`, `__repr__`, frozen guard, and slots) and just gate construction with a module-private sentinel:

```python
_FROM_YAML: Final = object()

@final
@dataclass(frozen=True, kw_only=True, slots=True, eq=False)
class PromptTemplate:
    name: str
    description: str | None
    parameter_keys: tuple[str, ...]
    _template: Template = field(repr=False)
    _token: InitVar[object]

    def __post_init__(self, _token: object) -> None:
        if _token is not _FROM_YAML:
            raise TypeError("Use PromptTemplate.from_yaml()")
```

`from_yaml` then just calls `cls(name=..., description=..., parameter_keys=tuple(definition.parameters), _template=_compile_template(definition, path=path), _token=_FROM_YAML)`. That deletes `_from_validated_definition`, both `object.*` calls, the custom raising `__init__`, and all four lint suppressions — for the same guarantees.

I checked this on 3.11: direct construction, bare `PromptTemplate()`, and a wrong/absent token all raise `TypeError`; instances stay frozen (`FrozenInstanceError`); equality stays identity-based; the repr still hides `_template`; and because `_token` is an `InitVar` it never lands in `__slots__` or on the instance. If you'd rather not carry a sentinel field, the other option is dropping `@dataclass` for a plain `@final` + `__slots__` class — but then you hand-roll frozen and `__repr__`, so this is the smaller change.

rampart/common/templates.py
spencrr: See #135

rampart/common/templates.py
spencrr: @nina-msft  Yeah I am not a big fan of the token/sentinel thing. Can you PTAL at #144 - maybe a cleaner approach?

nina-msft: If we go with #144 - can you close this out as stale?

spencrr: Closing both in favor of your comment! Stay tuned!

spencrr:
See #156

PR #135, opened 2026-07-27
[FIX]: Change ruff `noqa`s to human readable

spencrr:
## Description

Change all `noqa: XXX` -> `ruff: ignore[rule]` and bump ruff. Since `ruff check --fix` with preview now changes these codes, modifying comments to pass CI.

Note also, `extend-ignore` is deprecated, using `ignore`. And toml complaints about out-of-order tables in `pyproject.toml` fixed. `docs/**/*.md` also excluded from format (introduced with ruff bump).

## Breaking changes
None

## Checklist

- [ ] Documentation updated


pyproject.toml
```diff
@@ -39,7 +39,15 @@ dependencies = [
 onedrive = [
     "msgraph-sdk>=1.57.0",
     "azure-identity>=1.25.3"
- ]
+]
+
+[project.urls]
+Homepage = "https://github.com/microsoft/RAMPART"
+Repository = "https://github.com/microsoft/RAMPART"
```
bashirpartovi, line 46: What do you think if we use `aka.ms` links? e.g. aka.ms/RAMPART

bashirpartovi (approved):
LGTM, just a minor suggestion

pyproject.toml
spencrr: Ack - can change in follow up.

PR #146, opened 2026-08-04
[FEAT]: Add durable terminal evaluation contract

spencrr:
## Description

Adds the per-execution provenance needed before final-trace verdict cadence changes. `Result.terminal_evaluation` stores the evaluator output for the terminal trace, `Result.trace_end_reason` records why trace production ended, and `Turn.eval_purpose` identifies online stop checks. `Result.turn_evaluations` makes the online evidence boundary explicit while `eval_results` remains a compatibility view.

The direct attack and probe resolvers require one evaluation and reject unknown runtime outcomes instead of falling through. Population provenance now shares validation across `PopulationRef`, `PopulationResult`, and `execute_trials_async`, with invalid thresholds rejected before an execution factory runs.

This PR also includes the persistence work previously split into #147 so the contract cannot land without transport support. The current xdist v2 envelope and JSON report carry terminal evaluation, trace-end reason, turn purpose, and population provenance together. Malformed worker data fails closed, including overflowing confidence values, and oversized results produce bounded incomplete markers while retaining population provenance when it fits.

## Breaking changes

None for valid callers. Invalid population provenance and malformed evaluator outcomes now fail early instead of being accepted or falling through.

## Checklist

- [x] Tests cover contract fields, compatibility views, resolvers, population validation, xdist transport, truncation, JSON output, malformed confidence, and real worker/controller persistence
- [x] Documentation updated

Validation: 1,122 unit tests pass. Strict documentation build and all pre-commit checks pass.


docs/usage/results-and-reporting.md
```diff
@@ -49,9 +51,31 @@ for turn in result.turns:
...
+
+This layer makes terminal provenance durable before changing execution
+cadence. Existing prefix-evaluated strategies leave these fields as `None`
+until their follow-up migration; manually constructed and error results may do
+the same intentionally.
+
+Online evaluations attached to turns are available as
+`result.turn_evaluations`. The older `result.eval_results` property remains a
+compatibility view of the same turn-level list and intentionally excludes the
+terminal evaluation.
```
nina-msft, line 70: Since we aren't following an explicit deprecation strategy right now - can we remove this `eval_results` property and go directly with `turn_evaluations` if it won't be used in the future?

rampart/core/_population.py
```diff
@@ -0,0 +1,124 @@
...
+        str: Validated population identifier.
+
+    Raises:
+        TypeError: If ``value`` is not a string.
+        ValueError: If ``value`` is empty or exceeds the transport bound.
+    """
+    if not isinstance(value, str):
+        msg = "population id must be a string"
+        raise TypeError(msg)
+    if not value:
```
nina-msft, line 24: nit: Check for `if not value` before `if not isinstance(value, str)` since None will not be a string regardless :)

rampart/core/_population.py
```diff
@@ -0,0 +1,124 @@
...
+
+def validate_population_id(value: object) -> str:
+    """Validate and return a population identifier.
+
+    Returns:
+        str: Validated population identifier.
+
+    Raises:
+        TypeError: If ``value`` is not a string.
+        ValueError: If ``value`` is empty or exceeds the transport bound.
```
nina-msft, line 19: nit: only returns ValueError is is empty - I don't see any logic around a transport bound?

rampart/core/result.py
```diff
@@ -169,10 +204,19 @@ def safe(self) -> bool:
...
-        """Evaluator outcomes derived from turns."""
+    def turn_evaluations(self) -> list[EvalResult]:
+        """Online evaluator outcomes attached to turns."""
         return [t.eval_result for t in self.turns if t.eval_result is not None]
 
+    @property
+    def eval_results(self) -> list[EvalResult]:
+        """Compatibility view of online evaluations attached to turns.
+
+        ``terminal_evaluation`` is intentionally excluded. New consumers
```
nina-msft, line 215: same note as L67 in results-and-reporting.md, we should just remove this now since we are ok with breaking changes & fix any of the call sites to call turn_evaluations directly :)

rampart/core/result.py
```diff
@@ -287,6 +335,66 @@ def __repr__(self) -> str:
...
+
+    Returns:
+        SafetyStatus: DETECTED maps to UNSAFE, NOT_DETECTED maps to SAFE,
+            and UNDETERMINED is preserved.
+
+    Raises:
+        ValueError: If ``evaluation.outcome`` is not a known EvalOutcome.
+    """
+    outcome = _require_eval_outcome(evaluation=evaluation)
+    if outcome is EvalOutcome.DETECTED:
```
nina-msft, line 352: nit: stylistically i like not having many returns. i'd rewrite like

```
status = SafetyStatus.UNDETERMINED
if outcome is EvalOutcome.DETECTED:
    status = SafetyStatus.UNSAFE
if outcome is EvalOutcome.NOT_DETECTED:
    status = SafetyStatus.SAFE

return status
```

rampart/core/result.py
```diff
@@ -287,6 +335,66 @@ def __repr__(self) -> str:
...
+        evaluation: The terminal-trace evaluator result.
+
+    Returns:
+        SafetyStatus: DETECTED maps to SAFE, NOT_DETECTED maps to UNSAFE,
+            and UNDETERMINED is preserved.
+
+    Raises:
+        ValueError: If ``evaluation.outcome`` is not a known EvalOutcome.
+    """
+    outcome = _require_eval_outcome(evaluation=evaluation)
```
nina-msft, line 372: same note about many returns here - see L352

docs/usage/results-and-reporting.md
```diff
@@ -49,9 +51,31 @@ for turn in result.turns:
     turn.response.text        # What came back
     turn.response.tool_calls  # Tool invocations observed
     turn.eval_result          # EvalResult for this turn, or None
+    turn.eval_purpose         # EvaluationPurpose | None
     turn.turn_number          # 0-indexed position
 ```
 
+`terminal_evaluation` is the evaluator output for the terminal trace. It is an
```
nina-msft, line 58: Naming thoughts: Since this is the final-trace evaluator output vs turn-level evals...what do you think about naming this `final_trace_evaluation`? Terminal does not read to me as common of language as "trace" in our docs. It can even be `final_evaluation` or something without trace/terminal...but if we use trace then it would relate it to `trace_end_reason` while still adding clarity.

PR #152, opened 2026-08-05
[FEAT]: Snapshot collector results at call phase

nina-msft:
## Description

**PR 1 of 2** toward durable, per-test xdist result transport. This PR adds the *read seam* only — no transport wiring, no behavior change.

Today a test's `Result` objects are collected into a per-test `ResultCollector` (scoped by a ContextVar in the autouse `_rampart_collect` fixture) and only **absorbed at fixture teardown**. That is too late for a per-test transport: pytest-xdist ships its `TestReport` for a test at the **call phase**, before teardown runs, so a streaming transport would have nothing to read yet.

This PR makes the per-test results readable at call phase:

- Adds `get_active_collector()` to `_collection.py` — a thin accessor for the active-collector ContextVar.
- Adds a `pytest_runtest_makereport` **wrapper** hook that, on the `call` phase, snapshots `collector.results` onto `item.stash[_call_results_key]` while the collector is still active — before the fixture's teardown drains and deactivates it.
  - `wrapper=True` so it always runs and returns the report unchanged, regardless of `firstresult` registration order.
  - **Gated to xdist workers** (`is_xdist_worker`): single-process and controller runs never consume the snapshot, so they stay allocation-free and byte-identical to their pre-snapshot behavior.

The new stash key is an additive side-channel with **no consumer in this PR**, so there is no observable change to any existing run mode.

### How this sets up the follow-up PR

**PR 2 — Streamed transport** builds directly on this seam and needs no further lifecycle changes:

- **Worker side:** a call-phase hook reads `item.stash[_call_results_key]`, serializes each `Result` with the existing `_serialize_result`, and attaches it to the `TestReport` that execnet *already ships per test* (`report._rampart_results`).
- **Controller side:** `pytest_runtest_logreport` deserializes and merges each result incrementally into the session — replacing the current end-of-worker `workeroutput` **batch**.

Why this matters: the batch has a durability hole — if a worker dies mid-session, `workeroutput` is never written and **all** of that worker's results are lost. Streaming delivers each result the moment its test finishes, so pre-crash results survive. A de-risk spike measured ~1% memory overhead with 100% delivery through ~1 GiB.

Splitting the collector-lifecycle change into its own PR keeps it small and reviewable, and lets the fixture-affecting change (it runs on *every* test) land and bake before the transport builds on it — per the request on #113 not to fold this into that work.

Context: supersedes the closed-unmerged exploration in #113 (filesystem shards), which is not on `main`.

## Breaking changes

None. Additive only; the new stash key has no consumer, and the snapshot is skipped entirely outside xdist workers.

## Checklist

- [ ] Documentation updated

### DIFF
--- a/rampart/pytest_plugin/plugin.py
+++ b/rampart/pytest_plugin/plugin.py
@@ -41,6 +41,7 @@ from rampart.pytest_plugin._collection import (
     ResultCollector,
     activate_collector,
     deactivate_collector,
+    get_active_collector,
 )
 from rampart.pytest_plugin._session import RampartSession
 from rampart.pytest_plugin._xdist import (
@@ -69,14 +70,19 @@ __all__ = [
     "pytest_addoption",
     "pytest_collection_modifyitems",
     "pytest_configure",
+    "pytest_runtest_makereport",
     "pytest_sessionfinish",
     "pytest_terminal_summary",
     "pytest_testnodedown",
     "pytest_unconfigure",
 ]
 
+# Config-scoped stash keys: one entry per pytest session.
 _rampart_key = pytest.StashKey[RampartSession]()
 _session_start_key = pytest.StashKey[float]()
+
+# Item-scoped stash key: a per-test snapshot consumed by xdist streaming.
+_call_results_key = pytest.StashKey[list[Result]]()
 
 # Module-level constants are an acceptable exception in a hook-based
 # plugin module where there is no natural owning class.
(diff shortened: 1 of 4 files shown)


rampart/pytest_plugin/plugin.py
```diff
@@ -454,6 +460,42 @@ def _rampart_collect(  # pytest discovers this via autouse=True
         )
 
 
+@pytest.hookimpl(wrapper=True)
```
spencrr, line 463: Just making sure we don't need a trylast here

rampart/pytest_plugin/plugin.py
```diff
@@ -454,6 +460,42 @@ def _rampart_collect(  # pytest discovers this via autouse=True
...
+    Returns:
+        Generator[None, pytest.TestReport, pytest.TestReport]: The
+            unchanged report produced by downstream hookimpls.
+    """
+    report = yield
+    if call.when == "call" and is_xdist_worker(config=item.config):
+        collector = get_active_collector()
+        if collector is not None:
+            item.stash[_call_results_key] = collector.results
+    return report
```
spencrr, line 496: This is the meat of the pr so all we do is write to the stash of the specific item. Just making sure that `ResultCollector` will always be scoped to a single test. Specifically, is `_active_collector` ever used by concurrent running tests? I don't think so just checking.

nina-msft: Yep, correct on both. The collector is strictly per-test: the autouse fixture is function-scoped, so each test gets a brand-new `ResultCollector` set into the ContextVar on setup and cleared on teardown. Different tests never run at the same time in one process — sequential runs go one at a time, and xdist workers are separate processes, each with their own ContextVar. (Async tests also finish before the next starts.) The only overlap is child tasks *within* a single test, e.g. `asyncio.gather` — those intentionally share that test's collector via the copied context, which is exactly what we want.

rampart/pytest_plugin/plugin.py
nina-msft: `trylast` isn't needed bc this hook hands the report straight back (`report = yield` → `return report`) and never reads it, so where it runs relative to other `makereport` hooks doesn't matter. And the thing it *does* read — the collector — is fully populated by the end of the call phase and isn't cleared until teardown (a later phase), so it's ready no matter the ordering. The one guarantee we actually need is `wrapper=True`, which keeps the `firstresult` builtin from skipping us.

spencrr: Yep makes sense! just read this! https://docs.pytest.org/en/7.1.x/how-to/writing_hook_functions.html#hook-function-ordering-call-example

rampart/pytest_plugin/plugin.py
spencrr: Got it! Confused `ContextVar` with an atomic, overwrite-able variable!

spencrr (approved):
🚀

PR #156, opened 2026-08-06
[REFACTOR]: Decouple PromptTemplate creation from YAML

spencrr:
## Description

This supersedes the constructor approaches in #132 and #144 and follows the review feedback on #144.

This PR makes prompt-template construction source-neutral while retaining `PromptTemplate.from_yaml(path)` as the YAML and filesystem adapter.

- Adds a public, strict, frozen `PromptTemplateDefinition`.
- Normalizes declared parameters to an immutable, ordered tuple.
- Prevents prompt source from appearing in definition or template representations.
- Replaces injectable dataclass construction with `PromptTemplate(definition=...)`.
- Compiles and validates Jinja exclusively inside `PromptTemplate`.
- Preserves `from_yaml(path)` and both existing production call sites.
- Keeps YAML paths outside the definition and compiled runtime object.
- Preserves direct underlying exception causes while adding YAML path context.
- Reports Jinja line numbers explicitly as template-value lines rather than misleading physical YAML lines.
- Replaces metadata-based equality with object identity.

Jinja receives neither the YAML path nor the logical template name. This avoids incorrect YAML file/line attribution and prevents names such as `summary.html` from unexpectedly enabling autoescape.

## Breaking changes

Direct component-wise construction is no longer accepted:

```python
PromptTemplate(
    name="Greeting",
    description=None,
    parameter_keys=("subject",),
    _template=compiled_template,
)
```

Construct from a validated definition instead:

```python
PromptTemplate(
    definition=PromptTemplateDefinition(
        name="Greeting",
        description=None,
        parameters=("subject",),
        value="Hello, {{ subject }}!",
    ),
)
```

`PromptTemplate.from_yaml(path)` is unchanged, so existing supported call sites require no migration.

Separate `PromptTemplate` instances now use identity equality and hashing rather than metadata-based structural equality. `PromptTemplateDefinitionError.path` is optional for in-memory construction.

## Checklist

  - 34 focused prompt-template tests pass
  - 652 tracked unit tests pass
- [x] Documentation updated
  - Module and public API docstrings describe the source-neutral construction and error contracts

### DIFF
--- a/rampart/common/templates.py
+++ b/rampart/common/templates.py
@@ -1,27 +1,18 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-"""YAML prompt-template loader shared by drivers and evaluators.
+"""Source-neutral prompt-template definition, compilation, and YAML loading.
 
-RAMPART keeps prompt text in YAML files under per-package ``prompts/``
-directories. Each file uses this schema:
-
-- ``name`` (required): Human-readable template name.
-- ``parameters`` (required): Exact keyword names accepted by ``render()``.
-- ``value`` (required): Jinja2 template source.
-- ``description`` (optional): Human-readable purpose; defaults to ``None``.
-
-Unknown keys and type coercion are rejected. Parameter names must be unique.
-The loader maps ``parameters`` to :attr:`PromptTemplate.parameter_keys`,
-compiles ``value``, and returns a structured template that validates every
-render call before evaluating Jinja markup.
+``PromptTemplate`` accepts only a validated ``PromptTemplateDefinition`` and
+compiles its Jinja source internally. ``PromptTemplate.from_yaml`` is the
+filesystem and YAML adapter used by RAMPART's built-in drivers and evaluators.
+Every render call must provide exactly the definition's declared parameters.
 """
 
 from __future__ import annotations
 
 from collections.abc import Hashable
-from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Annotated, TypeAlias, TypeVar
+from typing import TYPE_CHECKING, Annotated, TypeAlias, TypeVar, final
 
 import yaml
 from jinja2 import (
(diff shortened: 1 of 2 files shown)


rampart/common/templates.py
```diff
@@ -248,26 +341,43 @@ def _compile_template(
...
+            details=exc.message or str(exc),
+            template_name=definition.name,
+            template_line=exc.lineno,
+        ) from exc
     except TemplateError as exc:
-        raise PromptTemplateDefinitionError(path=path, details=str(exc)) from exc
+        raise PromptTemplateDefinitionError(
+            details=str(exc),
+            template_name=definition.name,
+        ) from exc
```
nina-msft, line 356: Should both of these Error types list out `template_line`? Right now only the syntax error includes it. If they end up being the same, can we catch both and then raise the same type of error?

rampart/common/templates.py
```diff
@@ -248,26 +341,43 @@ def _compile_template(
...
+        raise PromptTemplateDefinitionError(
+            details=exc.message or str(exc),
+            template_name=definition.name,
+            template_line=exc.lineno,
+        ) from exc
     except TemplateError as exc:
-        raise PromptTemplateDefinitionError(path=path, details=str(exc)) from exc
+        raise PromptTemplateDefinitionError(
+            details=str(exc),
+            template_name=definition.name,
```
nina-msft, line 382: Same question here as comment above

nina-msft (approved):
This looks like a great implementation based off of my feedback from previous PR! I did not get a chance to run this through GHCP because it is being _very_ slow for me today - so if you haven't done a self-review through that it may be worth to make sure we didn't miss anything.

Otherwise, small feedback re: exception handling and otherwise I think the direction looks good. Up to you if you'd also like @bashirpartovi to take a look before merge.

rampart/common/templates.py
spencrr: https://jinja.palletsprojects.com/en/stable/api/#jinja2.TemplateError has no lineno, only `TemplateSyntaxError` 😄

PR #22, opened 2026-04-21
[CI]: Add automated linter tool for repo-level rules

spencrr:
## Description

Add `rampart_lint` — a custom AST-based linter that enforces two project coding conventions that ruff cannot check natively:

- **RAMPART001** — async functions must end with `_async` suffix (dunders exempt)
- **RAMPART002** — functions with >1 non-self/cls parameter must use keyword-only `*` (dunders exempt)

### Architecture

```
tools/rampart_lint/
  __main__.py           # CLI: argparse subcommands (check, list, explain)
  driver.py             # engine: AST walking, dispatch, noqa suppression
  rule.py               # Rule protocol, LocatedNode type alias, shared utilities
  rules/
    rampart001_async_suffix.py
    rampart002_keyword_only.py
  tests/                # co-located, 66 tests
```

Rules implement a `Rule` protocol and register concrete AST node types.

Suppression follows ruff semantics — same-line `# noqa: RAMPART001` or bare `# noqa`. Codes are declared as `external` in ruff config so `RUF100` won't flag them.

### Integration

- **pre-commit**: `rampart-lint` hook runs on rampart files
- **CI**: added `RAMPART lint` step to lint job; test job now uses `testpaths` from `pyproject.toml` (includes tool tests)
- **`pyproject.toml`**: `external`, `testpaths`, pyright `include`, per-file-ignores for tools

### Existing code

Added `# noqa: RAMPART002` to 3 pytest hook functions in plugin.py whose signatures are defined by the pytest framework.

## Breaking changes

None

## Checklist

- [x] `pre-commit run --all-files` passes
- [x] Tests added or updated for changes
- [x] Documentation updated

PR #158, opened 2026-08-08
[CI]: Add flake8 local plugin enforcing the async naming convention

spencrr:
## Description

Follow-up to #22 (closed unmerged), taking @nina-msft's suggestion to use a flake8 plugin instead of a custom linter. Covers the `_async` naming rule only.

69 lines instead of 811. No source files change.

flake8 supplies the noqa parser, CLI, file discovery and rule dispatch that #22 hand-built. What remains is the rule itself, ~9 lines.

Two deviations from the suggested snippet:

- **Prefix is `RMP`, not `RAMPART`.** flake8 validates codes against `^[A-Z]{1,3}[0-9]{0,3}$` and refuses to start otherwise.
- **No packaging.** [Local plugins](https://flake8.pycqa.org/en/latest/user/configuration.html#using-local-plugins) drop the entry point: `.flake8` points at `tools/`. Nothing ships in the `rampart` wheel.

On running two linters: ruff [has no plugin system](https://docs.astral.sh/ruff/faq/#does-ruff-support-plugins) and no rule for name suffixes. `select = RMP` means flake8 runs only this rule, so it never overlaps ruff.

### Baseline

159 pre-existing violations are recorded as `per-file-ignores` in `.flake8` rather than fixed here, keeping this diff reviewable. The rule is live now; follow-up commits shrink the baseline to nothing.

### Silent no-op guard

`flake8 --select=RMP` exits 0 if nothing owns the prefix. Mitigated two ways: any load failure is a hard error (exit 1), and 3 subprocess tests run the real `.flake8` config to assert the rule actually fires.

### Notes

- `RMP001` is in `[tool.ruff.lint] external` so `RUF102` accepts `# noqa: RMP001`. Split: `# ruff: ignore[rule]` for ruff, `# noqa:` for flake8.
- Pre-commit hook triggers on `.flake8` too, so config edits can't skip it and fail CI later.

## Breaking changes

None.

## Checklist

- [x] Tests added or updated for changes — 10 in `tests/unit/tools/test_flake8_rampart.py`: 7 unit (dunder exemption, nested/class-level defs, sync ignored, position) + 3 subprocess (registration, noqa, `select` scoping)
- [x] Documentation updated — Automated Enforcement section in `coding-standards.instructions.md`; `code-style.md` (which still said this rule needed "a human eye") and `development-setup.md`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
       types: [python]
       pass_filenames: false
     - id: flake8-rampart
-      name: flake8-rampart (RMP001)
+      name: flake8-rampart (RMPXXX)
       entry: uv run flake8
       language: system
       # flake8 scans its whole configured scope, so trigger on any Python file
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,3 +14,12 @@ repos:
       language: system
       types: [python]
       pass_filenames: false
+    - id: flake8-rampart
+      name: flake8-rampart (RMP001)
+      entry: uv run flake8
+      language: system
+      # flake8 scans its whole configured scope, so trigger on any Python file
+      # (mirroring the ty hook above) plus the config itself. `types` is omitted
+      # so that .flake8, which is not Python, still triggers the hook.
+      files: (\.py$|^\.flake8$)
+      pass_filenames: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
       types: [python]
       pass_filenames: false
     - id: flake8-rampart
-      name: flake8-rampart (RMPXXX)
+      name: flake8-rampart
       entry: uv run flake8
       language: system
       # flake8 scans its whole configured scope, so trigger on any Python file
--- a/docs/contributing/code-style.md
+++ b/docs/contributing/code-style.md
@@ -8,7 +8,7 @@ RAMPART enforces a consistent code style through automated tooling and documente
 |------|---------|-----------------|
 | [Ruff](https://docs.astral.sh/ruff/) | Linting and formatting | `pyproject.toml` `[tool.ruff.*]` |
 | [ty](https://github.com/astral-sh/ty) | Static type checking | `pyproject.toml` `[tool.ty]` |
-| [flake8](https://flake8.pycqa.org/) | The one project convention Ruff cannot express | `.flake8`, `tools/flake8_rampart.py` |
+| [flake8](https://flake8.pycqa.org/) | For project conventions Ruff cannot express | `.flake8`, `tools/flake8_rampart.py` |
 | [pre-commit](https://pre-commit.com/) | Git hooks for automated checks | `.pre-commit-config.yaml` |
 
 ### Running Checks
@@ -34,7 +34,7 @@ A few details worth knowing:
 
 - **Ruff** is configured with `select = ["ALL"]`. Test files have relaxed rules (no docstrings, no type annotations, magic values allowed) via `per-file-ignores` in `pyproject.toml`.
 - **ty** targets Python 3.11 — every function needs complete parameter and return type annotations.
-- **flake8** runs only `RMP001` (`select = RMP` in `.flake8`), so it never overlaps Ruff. Suppress it with `# noqa: RMP001`; Ruff rules use `# ruff: ignore[rule-name]`.
+- **flake8** runs only `RMP` codes (`select = RMP` in `.flake8`), so it never overlaps Ruff. Suppress an RMP code with `# noqa: <CODE>` (e.g., `# noqa: RMP001`); Ruff rules use `# ruff: ignore[rule-name]`.
 
 
 ## Key Conventions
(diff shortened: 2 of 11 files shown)


docs/contributing/code-style.md
```diff
@@ -60,7 +62,7 @@ async def send_request_async(self, *, payload: str) -> Response: ...
 async def send_request(self, *, payload: str) -> Response: ...
 ```
 
-Dunder methods (`__aenter__`, `__aexit__`) are exempt.
+Dunder methods (`__aenter__`, `__aexit__`) are exempt. Enforced by `RMP001`.
```
nina-msft, line 65: Is this true? I thought RMP001 was just for _async :)

docs/contributing/code-style.md
```diff
@@ -33,6 +34,7 @@ A few details worth knowing:
 
 - **Ruff** is configured with `select = ["ALL"]`. Test files have relaxed rules (no docstrings, no type annotations, magic values allowed) via `per-file-ignores` in `pyproject.toml`.
 - **ty** targets Python 3.11 — every function needs complete parameter and return type annotations.
+- **flake8** runs only `RMP001` (`select = RMP` in `.flake8`), so it never overlaps Ruff. Suppress it with `# noqa: RMP001`; Ruff rules use `# ruff: ignore[rule-name]`.
```
nina-msft: Should we make this documentation forward-looking and generalize that flake8 runs only `RMP` custom rules that do not overlap with ruff instead of calling out `RMP001` specifically, in case we add more later?

docs/contributing/code-style.md
```diff
@@ -8,11 +8,12 @@ RAMPART enforces a consistent code style through automated tooling and documente
 |------|---------|-----------------|
 | [Ruff](https://docs.astral.sh/ruff/) | Linting and formatting | `pyproject.toml` `[tool.ruff.*]` |
 | [ty](https://github.com/astral-sh/ty) | Static type checking | `pyproject.toml` `[tool.ty]` |
+| [flake8](https://flake8.pycqa.org/) | The one project convention Ruff cannot express | `.flake8`, `tools/flake8_rampart.py` |
```
nina-msft: Same as below comment - to future generalize should we make this "the project convention(s) ruff cannot express" vs calling out exact numbers of how many exist at the moment

docs/contributing/code-style.md
spencrr: Yep! We ignore dunders

docs/contributing/code-style.md
spencrr: Maybe - I don't see adding many more soon. Wdyt?

docs/contributing/code-style.md
```diff
@@ -157,11 +159,10 @@ Before committing, run pre-commit — it covers everything the automated tooling
 uv run pre-commit run --all-files
 ```
 
-This runs Ruff (linting + formatting) and ty (type checking), which together enforce the copyright header, type annotations, log formatting, import organization, and most other conventions on this page.
+This runs Ruff (linting + formatting), ty (type checking), and flake8 (`RMP001`), which together enforce the copyright header, type annotations, log formatting, import organization, the `_async` suffix, and most other conventions on this page.
```
nina-msft: same suggestion - don't call out RMP001 specifically but rather the RMP-prefixed rules at large

docs/contributing/code-style.md
spencrr: ~~Sorry! Don't follow the exact numbers part 😉~~

docs/contributing/code-style.md
nina-msft: oh lmao i just re-read this....ok great

docs/contributing/code-style.md
nina-msft: I think you can say "(e.g. `RMP001`)" in here to reference the concrete rule while still making the rest generalizable bc then we won't need to update this as we add more but I also am not super tied to this so do what you wish :P It is cheap to update as new rules are updated if we'd rather punt

.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
       types: [python]
       pass_filenames: false
     - id: flake8-rampart
-      name: flake8-rampart (RMP001)
+      name: flake8-rampart (RMPXXX)
```
nina-msft: does the name need to have (RMPXXX) in it? can it just be `flake8-rampart`?

docs/contributing/code-style.md
```diff
@@ -159,7 +159,7 @@ Before committing, run pre-commit — it covers everything the automated tooling
 uv run pre-commit run --all-files
 ```
 
-This runs Ruff (linting + formatting), ty (type checking), and flake8 (`RMP001`), which together enforce the copyright header, type annotations, log formatting, import organization, the `_async` suffix, and most other conventions on this page.
+This runs Ruff (linting + formatting), ty (type checking), and flake8 (`RMPXXX`), which together enforce the copyright header, type annotations, log formatting, import organization, the `_async` suffix, and most other conventions on this page.
```
nina-msft: nit: sorry, after seeing your other edits i like "flake8 (`RMP` codes)" better than `RMPXXX`

pyproject.toml
```diff
@@ -117,8 +117,8 @@ ignore = [
     "docstring-extraneous-exception", # Allow extraneous exceptions in docstrings to deliberately document the caller-visible contract, even if they propagate from a delegated callee.
 ]
 
-# RMP001 comes from the flake8 local plugin in tools/. Declaring it here stops
-# RUF102 (invalid-rule-code) from rejecting `# noqa: RMP001` as unknown.
+# RMPXXX comes from the flake8 local plugin in tools/. Declaring it here stops
```
nina-msft: same here, can we update this to "`RMP` codes come from...declaring _them_ here stops....from rejecting `# noqa: RMP001`, for example, as unknown."

nina-msft (approved):
yay!

PR #165, opened 2026-08-13
[PERF]: execute XPIA injection handle activation concurrently

apocalypse9949:
This PR improves XPIA attack setup performance by initializing injection handles concurrently.

## Performance: Concurrent XPIA Handle Initialization

`XPIAExecution._activate_handles_async` previously initialized injection handles sequentially. For attacks involving multiple injection surfaces, this caused independent asynchronous setup operations to execute serially.

This change uses `asyncio.gather(..., return_exceptions=True)` so independent setup operations such as remote file uploads proceed in parallel. A failing handle does not cancel siblings, which lets successful activations finish `enter_async_context` and register `__aexit__` cleanup before the first error is re-raised.

The existing readiness wait remains concurrent via `asyncio.TaskGroup`.

### Impact

Local testing showed approximately **45% lower activation time** for multi-surface scenarios compared with the sequential implementation.

## Testing

* `uv run pytest tests/unit/attacks/test_xpia.py -v` — passed
* Added a regression test where one handle fails and a delayed sibling still registers cleanup
* Verified existing test behavior and exception handling remain unchanged

## Files Changed

* `rampart/attacks/_xpia.py` — concurrent handle activation via non-cancelling `asyncio.gather`
* `tests/unit/attacks/test_xpia.py` — partial-failure cleanup coverage

### DIFF
--- a/rampart/attacks/_xpia.py
+++ b/rampart/attacks/_xpia.py
@@ -160,23 +160,26 @@ class XPIAExecution(BaseExecution):
     ) -> None:
         """Activate all injection handles and wait for readiness.
 
+        Handle ``__aenter__`` runs concurrently via ``asyncio.gather`` with
+        ``return_exceptions=True`` so a failing sibling does not cancel
+        others mid-entry.  That lets successful handles finish
+        ``enter_async_context`` and register ``__aexit__`` cleanup before
+        the first error is re-raised.  ``TaskGroup`` would cancel remaining
+        entries and could orphan remote payloads that were created but not
+        yet registered on the exit stack.
+
         Args:
             stack (AsyncExitStack): The exit stack managing cleanup.
         """
-
-        async def _enter_handle(h: InjectionHandle) -> None:
-            await stack.enter_async_context(h)
-
-        # Concurrent context entry (network uploads)
-        try:
-            async with asyncio.TaskGroup() as tg:
-                for handle in self._handles:
-                    tg.create_task(_enter_handle(handle))
-        except ExceptionGroup as eg:
-            # Unwrap the first exception for cleaner error reporting.
-            # BaseExecution already catches and reports exceptions,
-            # but ExceptionGroup obscures the underlying InfrastructureError.
-            raise eg.exceptions[0] from eg
+        # Concurrent context entry (network uploads); non-cancelling so
+        # successful siblings still register cleanup on the exit stack.
+        results = await asyncio.gather(
+            *(stack.enter_async_context(handle) for handle in self._handles),
+            return_exceptions=True,
+        )
+        errors = [result for result in results if isinstance(result, BaseException)]
+        if errors:
+            raise errors[0]
 
         # Concurrent readiness wait (indexing delays)
         async with asyncio.TaskGroup() as tg:
--- a/rampart/attacks/_xpia.py
+++ b/rampart/attacks/_xpia.py
@@ -163,10 +163,22 @@ class XPIAExecution(BaseExecution):
         Args:
             stack (AsyncExitStack): The exit stack managing cleanup.
         """
-        for handle in self._handles:
-            await stack.enter_async_context(handle)
 
-        # Concurrent: total = max of all wait times
+        async def _enter_handle(h: InjectionHandle) -> None:
+            await stack.enter_async_context(h)
+
+        # Concurrent context entry (network uploads)
+        try:
+            async with asyncio.TaskGroup() as tg:
+                for handle in self._handles:
+                    tg.create_task(_enter_handle(handle))
+        except ExceptionGroup as eg:
+            # Unwrap the first exception for cleaner error reporting.
+            # BaseExecution already catches and reports exceptions,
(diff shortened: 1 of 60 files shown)


apocalypse9949:

rampart/attacks/_xpia.py
```diff
@@ -163,10 +163,22 @@ async def _activate_handles_async(
...
+        # Concurrent context entry (network uploads)
+        try:
+            async with asyncio.TaskGroup() as tg:
+                for handle in self._handles:
+                    tg.create_task(_enter_handle(handle))
+        except ExceptionGroup as eg:
+            # Unwrap the first exception for cleaner error reporting.
+            # BaseExecution already catches and reports exceptions,
+            # but ExceptionGroup obscures the underlying InfrastructureError.
+            raise eg.exceptions[0] from eg
```
spencrr: we can also think about putting ExceptionGroup into a new InfrastructureError adjacent class but I think just unwrapping the first is likely sufficient!

rampart/pytest_plugin/_session.py
```diff
@@ -437,5 +443,6 @@ def build_report(self) -> TestRunReport:
             errors=counts[SafetyStatus.ERROR],
             duration_seconds=self._duration_seconds,
             metadata=metadata,
+            trial_groups=trial_groups_serialized,
```
spencrr: Thanks for bringing this up, @apocalypse9949 -- to remediate conflicts, though, I think it might be better for us to proceed with #121/#123 which is a bigger addition to how RAMPART handles trials. What do you think?

spencrr (commented):
Thanks for the suggestions, @apocalypse9949. I will still leave this open for comment from others 😄

rampart/pytest_plugin/_session.py
apocalypse9949: Makes sense. Since #121/#123 is a broader change to trial handling, I agree it would be cleaner to let those PRs handle the trial_groups changes and keep #165 focused on concurrent XPI injection handle activation. I'll avoid duplicating the trial-related changes here.

rampart/attacks/_xpia.py
apocalypse9949: Agreed. I think unwrapping the first exception is sufficient for now and keeps the change simple.

rampart/pytest_plugin/_session.py
nina-msft: @apocalypse9949 feel free to request another review by tagging us when you push a commit with these changes :-)

apocalypse9949: @nina-msft @spencrr  review pls I removed trial_group handlings

spencrr: LGTM - just need a couple linting fixes 😄

rampart/attacks/_xpia.py
```diff
@@ -163,10 +163,22 @@ async def _activate_handles_async(
         Args:
             stack (AsyncExitStack): The exit stack managing cleanup.
         """
-        for handle in self._handles:
-            await stack.enter_async_context(handle)
 
-        # Concurrent: total = max of all wait times
+        async def _enter_handle(h: InjectionHandle) -> None:
```
nina-msft: Linter flagged: async function `_enter_handle` must be named with an `_async` suffix (e.g. `_enter_handle_async`) to pass through CI

nina-msft:
Please update the PR description to match the scoped implementation :-)

rampart/attacks/_xpia.py
```diff
@@ -163,10 +163,22 @@ async def _activate_handles_async(
...
+            await stack.enter_async_context(h)
+
+        # Concurrent context entry (network uploads)
+        try:
+            async with asyncio.TaskGroup() as tg:
+                for handle in self._handles:
+                    tg.create_task(_enter_handle(handle))
```
nina-msft: `TaskGroup` cancels the remaining `__aenter__` calls as soon as one handle fails. A sibling may already have created its remote payload but be cancelled before `enter_async_context()` can register its `__aexit__`, leaving that payload orphaned.

Please replace `_enter_handle` and the first `TaskGroup` block with a non-cancelling gather; this belongs directly in `_activate_handles_async`, before the existing readiness-wait block:

```python
results = await asyncio.gather(
    *(stack.enter_async_context(handle) for handle in self._handles),
    return_exceptions=True,
)
errors = [result for result in results if isinstance(result, BaseException)]
if errors:
    raise errors[0]

# Keep the existing readiness TaskGroup here.
```

This preserves concurrent activation while allowing successful siblings to register cleanup. 

Please also add unit tests: a regression test where one handle fails immediately and another finishes activation after a delay would cover the partial-failure case.

nina-msft (changes requested):

rampart/attacks/_xpia.py
apocalypse9949: Done @nina-msft ready for review

nina-msft (approved):
Thanks for your contribution @apocalypse9949!

PR #166, opened 2026-08-13
[FEAT]: Stream xdist results through test reports

nina-msft:
## Description

The previous end-of-worker xdist batch could lose every Result from a worker that crashed before session finish and applied the size cap to the worker payload as a whole. This PR replaces that batch with incremental call-report streaming so the controller retains each Result as soon as its TestReport arrives.

This builds on the collector call-phase snapshot delivered by #152, which is now merged into `main`.

### Approach

- Workers attach an execnet-safe JSON envelope to call-phase TestReports using the existing full-fidelity Result serializer.
- The controller validates and merges envelopes in `pytest_runtest_logreport`, tags source workers, and preserves deterministic ordering, including `--dist=each`.
- Workeroutput contains only trial specifications and the expected streamed Result count. Node-down reconciliation marks missing output, missing counts, mismatches, and worker errors incomplete without discarding Results already received.
- The cap of 16 MiB applies independently to each Result. Oversized Results become attributed ERROR markers while normal Results continue to stream. A 4 KiB effective minimum and JSON-byte bounds guarantee markers fit, including control and non-BMP Unicode attribution.
- The transport schema and workeroutput key move to v2 so protocol skew fails closed.
- Results recorded only during fixture teardown remain intentionally outside the call-phase transport.

A killed worker retains every Result already delivered to the controller. The repository harness does not provide a deterministic remote/non-popen gateway, so that topology remains a residual integration risk.

### Validation

- `python -m pytest tests\unit\pytest_plugin -q` - 198 passed
- Focused streamed transport and real-xdist coverage - 95 passed
- Adversarial control/non-BMP cap coverage - 2 passed
- `ruff check .`
- `ruff format --check .`
- Repository-wide `ty check`

## Breaking changes

None. Mixed worker/controller transport versions remain unsupported and now fail closed under the v2 schema.

## Checklist

- [x] Tests added or updated for report streaming, async/call boundaries, slim workeroutput, reconciliation, crash durability, ordering, protocol skew, and per-Result caps
- [x] Documentation updated

### DIFF
--- a/rampart/pytest_plugin/_session.py
+++ b/rampart/pytest_plugin/_session.py
@@ -47,6 +47,34 @@ def _result_sort_key(result: Result) -> tuple[str, int, str]:
     return (nodeid, index, source_worker)
 
 
+def tag_collected_results(
+    *,
+    node: pytest.Item,
+    results: Sequence[Result],
+) -> list[Result]:
+    """Copy and tag Results with their pytest item metadata.
+
+    Returns:
+        list[Result]: Tagged shallow copies in their original order.
+    """
+    test_name = node.nodeid.split("::")[-1] if "::" in node.nodeid else node.nodeid
+    harm_marker = node.get_closest_marker("harm")
+    harm_category = harm_marker.args[0] if harm_marker and harm_marker.args else None
+    tagged: list[Result] = []
+    for result_index, original_result in enumerate(results):
+        result = copy.copy(original_result)
+        result.metadata = {
+            **result.metadata,
+            "_pytest_test_name": test_name,
+            "_pytest_nodeid": node.nodeid,
+            "_rampart_result_index": result_index,
+        }
+        if harm_category is not None and result.harm_category is None:
+            result.harm_category = harm_category
+        tagged.append(result)
+    return tagged
+
+
 @dataclass(frozen=True, kw_only=True)
 class TrialSpec:
     """Trial-clone metadata captured at collection time.
@@ -203,27 +231,7 @@ class RampartSession:
             node (pytest.Item): The test item that just completed.
             collector (ResultCollector): The test's result collector.
         """
-        test_name = node.nodeid.split("::")[-1] if "::" in node.nodeid else node.nodeid
-        harm_marker = node.get_closest_marker("harm")
-        harm_category = (
-            harm_marker.args[0] if harm_marker and harm_marker.args else None
-        )
-
-        collected = collector.results
-        tagged: list[Result] = []
-        for result_index, original_result in enumerate(collected):
-            # Shallow copy is sufficient because we reconstruct all
-            # mutable fields we modify (currently metadata and harm_category).
-            result = copy.copy(original_result)
-            result.metadata = {
-                **result.metadata,
-                "_pytest_test_name": test_name,
-                "_pytest_nodeid": node.nodeid,
-                "_rampart_result_index": result_index,
-            }
-            if harm_category is not None and result.harm_category is None:
-                result.harm_category = harm_category
-            tagged.append(result)
+        tagged = tag_collected_results(node=node, results=collector.results)
         self._results.extend(tagged)
         self._results_by_nodeid[node.nodeid] = tagged
         self._cached_report = None
@@ -385,7 +393,8 @@ class RampartSession:
                 in the report metadata.
         """
         self._incomplete = True
-        self._incomplete_reasons.append(reason)
+        if reason not in self._incomplete_reasons:
+            self._incomplete_reasons.append(reason)
         self._cached_report = None
 
     def set_report_metadata(self, *, metadata: dict[str, object]) -> None:
--- a/rampart/pytest_plugin/_session.py
+++ b/rampart/pytest_plugin/_session.py
@@ -57,7 +57,7 @@ def tag_collected_results(
     Returns:
         list[Result]: Tagged shallow copies in their original order.
     """
-    test_name = node.nodeid.split("::")[-1] if "::" in node.nodeid else node.nodeid
+    test_name = node.name
     harm_marker = node.get_closest_marker("harm")
     harm_category = harm_marker.args[0] if harm_marker and harm_marker.args else None
     tagged: list[Result] = []
--- a/rampart/pytest_plugin/_xdist.py
+++ b/rampart/pytest_plugin/_xdist.py
@@ -4,10 +4,9 @@
 """xdist support for RAMPART's pytest plugin.
 
 Provides serialization, deserialization, and controller-side merge
-logic for running RAMPART under pytest-xdist. Workers serialize their
-``Result`` objects into ``config.workeroutput``; the controller merges
-worker payloads in ``pytest_testnodedown`` and emits a single unified
-report at session end.
+logic for running RAMPART under pytest-xdist. Workers stream ``Result``
+objects on call-phase test reports; the controller merges each report
+incrementally and emits a single unified report at session end.
 
 Trust boundary: worker payloads may contain attacker-controlled
 content (agent responses, payload text). Serialization is strictly
@@ -57,13 +56,18 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-SCHEMA_VERSION: str = "rampart.xdist.v1"
-WORKEROUTPUT_KEY: str = "rampart_xdist_v1"
+SCHEMA_VERSION: str = "rampart.xdist.v2"
+WORKEROUTPUT_KEY: str = "rampart_xdist_v2"
+REPORT_RESULTS_ATTR: str = "_rampart_results"
 SIZE_LIMIT_OPTION: str = "rampart_xdist_max_bytes"
 DEFAULT_SIZE_LIMIT_BYTES: int = 64 * 1024 * 1024
+MIN_RESULT_SIZE_LIMIT_BYTES: int = 4 * 1024
 MAX_METADATA_DEPTH: int = 6
 
 _TRUNCATED_MARKER: str = "rampart_truncated"
+_TRUNCATED_ATTRIBUTION_MAX_BYTES: int = 512
+_TRUNCATED_FALLBACK_ATTRIBUTION_MAX_BYTES: int = 64
+_STREAMED_RESULT_COUNT: str = "streamed_result_count"
 
 
 class WorkerOutputError(Exception):
@@ -75,7 +79,7 @@ class SchemaVersionError(WorkerOutputError):
 
 
 class SizeLimitError(WorkerOutputError):
-    """Raised when a worker payload exceeds the configured size cap."""
+    """Raised when one serialized Result exceeds the configured size cap."""
 
 
 def is_xdist_worker(*, config: pytest.Config) -> bool:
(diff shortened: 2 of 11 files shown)


rampart/pytest_plugin/_session.py
```diff
@@ -47,6 +47,34 @@ def _result_sort_key(result: Result) -> tuple[str, int, str]:
...
+    *,
+    node: pytest.Item,
+    results: Sequence[Result],
+) -> list[Result]:
+    """Copy and tag Results with their pytest item metadata.
+
+    Returns:
+        list[Result]: Tagged shallow copies in their original order.
+    """
+    test_name = node.nodeid.split("::")[-1] if "::" in node.nodeid else node.nodeid
```
spencrr: I know it's just copied over to this new func, but can we use node.name or node.originalname instead of parsing here?

rampart/pytest_plugin/_xdist.py
```diff
@@ -498,29 +511,237 @@ def _serialize_result(*, result: Result, nodeid: str) -> dict[str, Any]:
...
+    nodeid: str,
+) -> dict[str, Any]:
+    """Serialize one Result, replacing only that Result when oversized.
+
+    Returns:
+        dict[str, Any]: Full Result data or a bounded truncation marker.
+    """
+    data = _serialize_result(result=result, nodeid=nodeid)
+    size_bytes = _serialized_size(data=data)
+    limit_bytes = _size_limit(config=config)
```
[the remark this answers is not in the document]
spencrr: good catch - maybe we can put this in a global for now? I think a follow up for me is to move most of these hooks (minus config which happens super early) to a Plugin class (that then we can have state in - also like a `is_worker` and controller cached property)

rampart/pytest_plugin/_xdist.py
```diff
@@ -498,29 +511,237 @@ def _serialize_result(*, result: Result, nodeid: str) -> dict[str, Any]:
...
+        int: Number of Result representations attached.
+    """
+    if not results:
+        return 0
+    data = serialize_report_data(
+        config=config,
+        nodeid=report.nodeid,
+        results=results,
+    )
+    setattr(report, REPORT_RESULTS_ATTR, data)
```
spencrr, line 718: this LGTM 😄

rampart/pytest_plugin/_xdist.py
```diff
@@ -944,48 +1165,59 @@ def _deserialize_result(*, data: object) -> Result:
...
+    deserialized: list[Result] = []
+    truncated = False
+    for index, raw_result in enumerate(cast("list[Any]", raw_results)):
+        result = _deserialize_result(data=raw_result)
+        result.metadata["_pytest_nodeid"] = nodeid
+        result.metadata["_rampart_result_index"] = index
+        deserialized.append(result)
+        if isinstance(raw_result, dict) and raw_result.get(_TRUNCATED_MARKER) is True:
+            truncated = True
+    return {nodeid: deserialized}, truncated
```
spencrr, line 1220: serde lgtm too! quick q on 1215-1216: is it required? seems same as in `_session.py::tag_collected_results`

> ```python
>         result.metadata = {
>             **result.metadata,
>             "_pytest_test_name": test_name,
>             "_pytest_nodeid": node.nodeid,
>             "_rampart_result_index": result_index,
>         }
> ```

spencrr, line 1220: also maybe not necessary maybe consider a small dataclass as return result over the tuple of a dict and bool?

rampart/pytest_plugin/_xdist.py
```diff
@@ -1053,38 +1290,19 @@ def finalize_worker(*, config: pytest.Config, session: RampartSession) -> None:
...
-            f"truncated. Increase --{SIZE_LIMIT_OPTION.replace('_', '-')} "
-            f"(or the {SIZE_LIMIT_OPTION} ini option) to raise the cap."
-        )
-        raise SizeLimitError(msg)
-    logger.debug("Worker payload size: %d bytes", size)
-    workeroutput[WORKEROUTPUT_KEY] = payload
+    workeroutput[WORKEROUTPUT_KEY] = serialize_worker_data(
+        session=session,
+        streamed_result_count=streamed_result_count,
+    )
```
spencrr, line 1305: looks SO much cleaner! 🧼

rampart/pytest_plugin/_xdist.py
```diff
@@ -1136,13 +1354,88 @@ def _tag_source_worker(
             result.metadata["_rampart_source_worker"] = worker_id_str
 
 
+def get_worker_id(node: object) -> str:
+    """Return xdist's stable identifier for a worker node."""
+    gateway = getattr(node, "gateway", None)
+    return str(getattr(gateway, "id", node)) if gateway else str(node)
+
```
spencrr, line 1361: https://pytest-xdist.readthedocs.io/en/latest/how-to.html#envvar-PYTEST_XDIST_WORKER

rampart/pytest_plugin/_xdist.py
```diff
@@ -1136,13 +1354,88 @@ def _tag_source_worker(
...
+) -> tuple[str, int] | None:
+    """Validate and incrementally merge one streamed report envelope.
+
+    Returns:
+        tuple[str, int] | None: Source worker and merged Result count,
+            or None when the report carries no RAMPART envelope.
+    """
+    payload = getattr(report, REPORT_RESULTS_ATTR, None)
+    if payload is None:
+        return None
```
spencrr, line 1395: failing silently here if no envelope - just making sure this is intended for non-RAMPART tests running in RAMPART suite? or throw warning?

rampart/pytest_plugin/_xdist.py
```diff
@@ -1136,13 +1354,88 @@ def _tag_source_worker(
...
+
+    Returns:
+        int: Expected streamed Result count.
+
+    Raises:
+        WorkerOutputError: If the count is missing or invalid.
+    """
+    typed = _validate_schema(data=payload)
+    raw_count = typed.get(_STREAMED_RESULT_COUNT)
+    if type(raw_count) is not int or raw_count < 0:
```
spencrr, line 1425: what does `type(...) is not int` vs `isinstance(int ...` do?

rampart/pytest_plugin/_xdist.py
```diff
@@ -1136,13 +1354,88 @@ def _tag_source_worker(
...
+def _deserialize_streamed_result_count(*, payload: object) -> int:
+    """Read the expected Result count from slim workeroutput.
+
+    Returns:
+        int: Expected streamed Result count.
+
+    Raises:
+        WorkerOutputError: If the count is missing or invalid.
+    """
+    typed = _validate_schema(data=payload)
```
spencrr, line 1423: we are validating twice here and `deserialize_report_data`?

rampart/pytest_plugin/plugin.py
```diff
@@ -493,9 +499,60 @@ def pytest_runtest_makereport(
...
+    """Merge a streamed xdist Result envelope on the controller.
+
+    Malformed envelopes mark the run incomplete instead of aborting
+    pytest, matching the workeroutput trust-boundary convention.
+
+    Args:
+        report (pytest.TestReport): A setup, call, or teardown report.
+    """
+    node = getattr(report, "node", None)
+    config = getattr(node, "config", None)
```
spencrr, line 530: I am not sure report.node.config works

rampart/pytest_plugin/_session.py
nina-msft: Fixed in `e9300d7`. `tag_collected_results` now uses `node.name` instead of parsing `node.nodeid`. I kept `node.name` rather than `originalname` so parameterized display identity is preserved, with a focused regression test.

rampart/pytest_plugin/_xdist.py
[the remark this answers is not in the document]
nina-msft: Fixed in `e9300d7`. The size limit is now resolved once per report envelope in `serialize_report_data` and passed into each `_serialize_capped_result` call. This avoids repeated config lookups and duplicate warnings without introducing global state. I also added coverage for a multi-Result envelope to verify one lookup/warning.

rampart/pytest_plugin/_xdist.py
nina-msft: 🤠 yeehaw!

rampart/pytest_plugin/_xdist.py
nina-msft: Yes, this is required. `tag_collected_results` supplies worker-side metadata, but the controller treats the transported Result metadata as untrusted. It reapplies node ID and result index from the validated envelope and list position so transported metadata cannot alter attribution or ordering.

nina-msft: A named return type could help if this grows. Currently the tuple has two values and only one production call site, so I’d keep it simple for now - wdyt?

rampart/pytest_plugin/_xdist.py
nina-msft: 🛁 🚿 ✨

rampart/pytest_plugin/_xdist.py
nina-msft: `PYTEST_XDIST_WORKER` is available inside the worker process, while these identity helpers are used controller-side. The streamed report carries xdist’s `worker_id`; the controller node’s gateway ID is only a fallback and is also used during node-down reconciliation.

rampart/pytest_plugin/_xdist.py
nina-msft: Returning `None` is intentional. RAMPART attaches an envelope only to a call report with at least one Result. Setup/teardown reports and tests recording no Results legitimately have no envelope, so warning here would flag normal execution.

rampart/pytest_plugin/_xdist.py
nina-msft: The exact type check intentionally rejects booleans. Because `bool` subclasses `int`, `isinstance(True, int)` is true, but `True` should not be accepted as a streamed Result count.

rampart/pytest_plugin/_xdist.py
nina-msft: There are two cases here. The streamed report envelope and slim workeroutput are separate boundary payloads, so each requires schema validation. Within node-down, the same slim workeroutput is also validated once by the trial-spec parser and once by the count parser. That duplication is intentional to keep both parsers independently safe and callable; it is only a small dictionary/version check.

rampart/pytest_plugin/plugin.py
nina-msft: xdist’s `DSession.worker_testreport` explicitly assigns `rep.node = node` before invoking `pytest_runtest_logreport`, and that `WorkerController` carries `config`. The real-xdist integration tests exercise this exact path.

rampart/pytest_plugin/_xdist.py
spencrr: Thanks, I traced this further. I agree these lines are required in the current implementation, but for a different reason: the oversized-result marker truncates `_pytest_nodeid` and omits `_rampart_result_index`, so controller-side reconstruction restores full attribution and ordering.

rampart/pytest_plugin/plugin.py
spencrr: ahh weird that they did it that way - it's not on the base TestReport. Got it!

spencrr (approved):
If it fits, it ships! 📦🚀

PR #173, opened 2026-08-24
[FIX]: Avoid eager PyRIT imports during pytest startup

spencrr:
## Description

Avoid eagerly importing PyRIT and Transformers when pytest loads the RAMPART plugin.

- Lazily resolve PyRIT-backed top-level exports using PEP 562.
- Preserve the existing public API and static type information.
- Add regression tests for lightweight plugin startup and lazy export caching.

Performance:
- Plugin import: 15.77s -> 0.49s
- Representative pytester tests: 48.50s -> 9.66s
- All 19 pytester tests complete in 181.56 ->19.01s

Validation:
- `pre-commit run --all-files`
- Full unit suite: 714 passed

## Breaking changes

None.

## Checklist

- [x] Documentation updated (not required; no user-facing API change)

### DIFF
--- a/rampart/__init__.py
+++ b/rampart/__init__.py
@@ -6,7 +6,9 @@
 Public API re-exports for convenient top-level access.
 """
 
-from rampart.attacks import Attacks
+from importlib import import_module
+from typing import TYPE_CHECKING
+
 from rampart.core.adapter import AgentAdapter, Session
 from rampart.core.errors import DriverError, EvaluatorError, InfrastructureError
 from rampart.core.evaluator import BaseEvaluator, Evaluator
@@ -41,10 +43,21 @@ from rampart.core.types import (
     ToolCall,
     Turn,
 )
-from rampart.drivers.llm import LLMDriver
-from rampart.evaluators import LLMJudge, TranscriptScope
-from rampart.probes import Probes
 from rampart.pytest_plugin._collection import record_result
+
+if TYPE_CHECKING:
+    from rampart.attacks import Attacks
+    from rampart.drivers.llm import LLMDriver
+    from rampart.evaluators import LLMJudge, TranscriptScope
+    from rampart.probes import Probes
+
+__lazy_imports__: dict[str, tuple[str, str]] = {
+    "Attacks": ("rampart.attacks", "Attacks"),
+    "LLMDriver": ("rampart.drivers.llm", "LLMDriver"),
+    "LLMJudge": ("rampart.evaluators", "LLMJudge"),
+    "Probes": ("rampart.probes", "Probes"),
+    "TranscriptScope": ("rampart.evaluators", "TranscriptScope"),
+}
 
 __all__ = [
     "AgentAdapter",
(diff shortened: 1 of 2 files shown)


rampart/__init__.py
```diff
@@ -41,11 +43,22 @@
...
-from rampart.probes import Probes
 from rampart.pytest_plugin._collection import record_result
 
+if TYPE_CHECKING:
+    from rampart.attacks import Attacks
+    from rampart.drivers.llm import LLMDriver
+    from rampart.evaluators import LLMJudge, TranscriptScope
+    from rampart.probes import Probes
+
+__lazy_imports__: dict[str, tuple[str, str]] = {
```
nina-msft, line 54: nit: suggest a private constant like `_LAZY_IMPORTS` instead since dunder names are reserved by python

nina-msft, line 54: note: the lazy export set now lives in three places — this dict, the `TYPE_CHECKING` block above, and `__all__` below. Adding a future heavy export means remembering to touch all three or the deferral silently breaks. Worth a short comment here noting they must stay in sync.

rampart/__init__.py
```diff
@@ -41,11 +43,22 @@
...
+    "Attacks": ("rampart.attacks", "Attacks"),
+    "LLMDriver": ("rampart.drivers.llm", "LLMDriver"),
+    "LLMJudge": ("rampart.evaluators", "LLMJudge"),
+    "Probes": ("rampart.probes", "Probes"),
+    "TranscriptScope": ("rampart.evaluators", "TranscriptScope"),
+}
+
 __all__ = [
```
nina-msft, line 62: caveat: these deferred names are still in `__all__`, so `from rampart import *` will re-trigger the full PyRIT import chain. That's fine for the pytest-startup goal we're fixing here, just noting the deferral isn't a complete escape hatch for star-import consumers.

rampart/__init__.py
spencrr: Asked AI to judge my helper-module proposal (since can't have private constant in init): 

1. RUF067 explicitly permits dunder assignments in `__init__.py`; `__lazy_imports__` is sanctioned, not a loophole.
2. A helper module only synchronizes the mapping with `__all__`; `TYPE_CHECKING` remains independently maintained.
3. It adds a file and eager import while weakening Ruff’s RUF022 validation of the explicit, sorted `__all__`.

Recommendation:

1. Keep the in-file `__lazy_imports__`.
2. Keep `__all__` explicit and sorted.
3. Add an invariant asserting lazy names are in `__all__`.

spencrr: See above comment about invariant

rampart/__init__.py
spencrr: `from rampart import *` should import all public names under `__all__`

rampart/__init__.py
nina-msft: ah ok didn't realize we can't have private consts in init funcs

nina-msft (approved):

PR #174, opened 2026-08-24
[CI]: Enforce lazy export synchronization

spencrr:
## Description

Follow-up to #173.

Add a repository-specific Flake8 rule that keeps PEP 562 lazy export registries synchronized with the public module API.

- Add `RMP002` to require every `__lazy_imports__` key in `__all__`.
- Require each declaration to use one direct, statically verifiable literal assignment.
- Reject redeclarations, additional writes, and direct calls rooted at `__all__` or `__lazy_imports__`; allow pure reads.
- Split the local plugin into independently registered `AsyncSuffixChecker` (`RMP001`) and `LazyExportChecker` (`RMP002`).
- Add focused AST tests for mixed registries, dynamic declarations, incremental construction, reads, and Flake8 wiring.
- Document the convention and register `RMP002` as an external Ruff code.

Validation:
- `python -m pre_commit run --all-files`
- Full unit suite: 732 passed

## Breaking changes

None.

## Checklist

- [x] Documentation updated

### DIFF
--- a/tests/unit/tools/test_flake8_rampart.py
+++ b/tests/unit/tools/test_flake8_rampart.py
@@ -85,6 +85,22 @@ __all__ = ["Eager", "Heavy"]
 
         assert _lazy_export_messages(source) == []
 
+    def test_reports_only_lazy_exports_missing_from_all(self) -> None:
+        source = """
+__lazy_imports__ = {
+    "Included": ("package.included", "Included"),
+    "MissingOne": ("package.missing", "MissingOne"),
+    "MissingTwo": ("package.missing", "MissingTwo"),
+}
+__all__ = ["Included"]
+"""
+
+        messages = _lazy_export_messages(source)
+
+        assert len(messages) == 2
+        assert "`MissingOne`" in messages[0]
+        assert "`MissingTwo`" in messages[1]
+
     def test_flags_lazy_export_when_all_is_absent(self) -> None:
         source = '__lazy_imports__ = {"Heavy": ("package.heavy", "Heavy")}'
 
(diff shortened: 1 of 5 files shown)


tools/flake8_rampart.py
```diff
@@ -34,11 +42,119 @@ def _is_dunder(name: str) -> bool:
...
+                isinstance(target, ast.Name) and target.id == name
+                for target in statement.targets
+            )
+        ) or (
+            isinstance(statement, ast.AnnAssign)
+            and isinstance(statement.target, ast.Name)
+            and statement.target.id == name
+        ):
+            value = statement.value
+    return value
```
nina-msft: > _This comment was generated by Copilot._

**Likelihood: low.** Today `__all__` is only ever built as a single literal assignment, so this can't misfire right now. It becomes a real risk only if someone later constructs `__all__` (or `__lazy_imports__`) incrementally — which is a fairly common pattern, so worth guarding before it can silently bite.

**Finding:** `_module_assignment` only tracks `Assign`/`AnnAssign` and keeps the last one, so it silently mis-analyzes augmented construction of these names:

```python
__all__ = ["Other"]
__all__ += ["Heavy"]        # or __all__.extend([...]) / .append(...)
```

Here it captures only `{"Other"}` and then flags `Heavy` as missing — a false positive. The concern is the asymmetry: fully dynamic forms are rejected *loudly* (`RMP002_DYNAMIC_ALL` / `RMP002_DYNAMIC_REGISTRY`), but `+=` / `.extend()` / `.append()` on `__all__` or `__lazy_imports__` is miscounted *silently*. Since the checker runs repo-wide, I'd rather it fail loud than wrong.

Consider: detect an `AugAssign` — and an attribute-call mutation (`.extend`/`.append`/`.update`) — targeting either name and treat it as dynamic, yielding the existing "must be a literal" violation.

tests/unit/tools/test_flake8_rampart.py
```diff
@@ -11,45 +11,123 @@
...
+}
+__all__ = []
+"""
+
+        (message,) = _lazy_export_messages(source)
+
+        assert message.startswith("RMP002")
+        assert "`Heavy`" in message
+
+    def test_accepts_lazy_export_in_all(self) -> None:
```
nina-msft, line 72: > _This comment was generated by Copilot._

**Likelihood: low** — these are coverage nice-to-haves, not a sign of a current bug; the existing tests already exercise the main paths. Worth adding mainly to pin down intent against future regressions.

Two cases worth adding:

1. Mixed registry — several lazy keys where only some are in `__all__`; assert exactly the missing name(s) are reported (guards against an all-or-nothing regression).
2. The `__all__ += [...]` / `.extend(...)` augmented form, asserting whatever behavior you settle on above — so the intent is documented either way.

nina-msft (commented):
Both coverage nits flagged by Copilot - lmk what you think!

tools/flake8_rampart.py
spencrr: `RMP002` now requires a single direct literal assignment and structurally rejects subsequent writes or direct calls rooted at either governed name. This makes `+=`, item assignment, and `.extend(...)` fail with the literal-declaration diagnostic instead of producing a misleading missing-export result. Pure reads remain allowed.

tests/unit/tools/test_flake8_rampart.py
spencrr: The mixed-registry test verifies that only omitted lazy names are reported. The incremental-construction tests cover representative AST shapes for augmented assignment, a rooted method call, and subscript assignment.

nina-msft (approved):
🚀 ✨

PR #175, opened 2026-08-25
[BREAKING] [MAINT]: remove rampart_sinks fixture and deprecation warnings

nina-msft:
## Description

Team discussed that formal deprecation cycles are unnecessary this early in the project, so this removes the deprecated `rampart_sinks` fixture path entirely rather than keeping it around with warnings. The `pytest_rampart_sinks` hook becomes the single, enforced way to register report sinks.

Key changes:

- Delete `rampart/common/deprecation.py` and its `emit_deprecation_warning` helper.
- Remove the `_rampart_sink_bootstrap` autouse fixture and the now-unused `_has_sink_hook_impl` guard from the pytest plugin; sink resolution now goes through the hook only, in both single-process and xdist controller runs.
- Remove `discover_sinks_from_conftest` and its fixture-unwrap helpers from `_xdist.py`.
- Add an explicit `--collect-only` guard in `pytest_sessionfinish`. "No reports on collect-only" was previously an accidental side effect of fixture timing; with the hook resolved unconditionally at session finish, it is now enforced directly so collection never overwrites report files.
- Migrate the xdist aggregation test conftest from the fixture to the hook (keeps end-to-end emission coverage) and drop the deprecation-contract tests.
- Update docs to describe hook-only sink registration.

Validation: `ruff check .` is clean and the full suite passes (1013 passed, 5 skipped), including the slow subprocess xdist tests.

## Breaking changes

Yes. Projects that register sinks via a `rampart_sinks` fixture will silently stop emitting reports. Migration: define a `pytest_rampart_sinks(config)` hook in `conftest.py` returning the same list of sinks.

```python
def pytest_rampart_sinks(config):
    return [JsonFileReportSink(output_dir=Path(".report"))]
```

## Checklist

- [x] Documentation updated


rampart/common/deprecation.py
```diff
@@ -1,56 +0,0 @@
-# Copyright (c) Microsoft Corporation.
```
nina-msft, line 1: Not sure if we want to keep this around in case we want to deprecate in the future. Since its only one file, we can easily write this up again once we decide this project needs a formal deprecation strategy

spencrr (approved):

PR #179, opened 2026-08-27
[FIX]: Preserve reports with duplicate timestamps

rioyu123:
## Description

`JsonFileReportSink` used a seconds-precision timestamp with `write_text()`. Two reports emitted within the same second therefore resolved to the same path, allowing the later report to silently replace the first.

This change preserves every report while keeping the common filename concise:

- Report filenames now use a UTC timestamp with millisecond precision, for example `run_report_2026-04-25T14-30-00-123.json`.
- The timestamp path is created exclusively with mode `x`, so an existing report is never overwritten.
- A UUID suffix is added only when the timestamp path already exists; the fallback path is also created exclusively.
- Reports continue to be fully serialized before any output file is opened, so serialization failures do not leave empty report files.
- Report content is written with explicit UTF-8 encoding instead of the platform default.

The usage guide and API documentation now describe the concise common filename, collision fallback, and remaining error cases. Tests cover repeated timestamps, existing-file preservation, a forced fallback collision, and serialization failure.

Validation:

- `uv run pre-commit run --all-files` — passed
- `uv run pytest tests/unit/reporting -q` — 46 passed
- `uv run pytest tests/unit -q` — 1035 passed on Linux; 1033 passed and 2 skipped on Windows
- `uv run --group docs mkdocs build --strict` — passed on Linux

## Breaking changes

There are no public API or report-schema changes. Generated filenames now include milliseconds and may include a UUID suffix on collision. Consumers using `run_report_*.json` are unaffected; consumers parsing the exact timestamp format should account for the additional millisecond segment.

## Checklist

- [x] Documentation updated

### DIFF
--- a/docs/usage/results-and-reporting.md
+++ b/docs/usage/results-and-reporting.md
@@ -87,9 +87,9 @@ from rampart.reporting import JsonFileReportSink
 sink = JsonFileReportSink(output_dir=Path(".report"))
 ```
 
-Output: `.report/run_report_2026-04-25T14-30-00-123_a3f18c92654d4b75ad15687d383d951b.json`
+Output: `.report/run_report_2026-04-25T14-30-00-123.json`
 
-The filename contains a UTC timestamp (millisecond precision) and a random UUID. Reports created in the same millisecond receive different filenames. An exact filename collision raises `FileExistsError` instead of overwriting an existing report. Reports written within the same millisecond have no defined filename order relative to each other.
+The filename contains a UTC timestamp with millisecond precision. If another report already has the same timestamp, a random UUID is appended to the new filename. Files are created atomically and existing reports are never overwritten. Reports written within the same millisecond have no defined filename order relative to each other.
 
 ### Custom Sinks
 
--- a/docs/usage/results-and-reporting.md
+++ b/docs/usage/results-and-reporting.md
@@ -87,7 +87,9 @@ from rampart.reporting import JsonFileReportSink
 sink = JsonFileReportSink(output_dir=Path(".report"))
 ```
 
-Output: `.report/run_report_2026-04-25T14-30-00.json`
+Output: `.report/run_report_2026-04-25T14-30-00-123_a3f18c92654d4b75ad15687d383d951b.json`
+
+The filename contains a UTC timestamp (millisecond precision) and a random UUID. Reports created in the same millisecond receive different filenames. An exact filename collision raises `FileExistsError` instead of overwriting an existing report. Reports written within the same millisecond have no defined filename order relative to each other.
 
 ### Custom Sinks
 
--- a/rampart/reporting/json_file.py
+++ b/rampart/reporting/json_file.py
@@ -31,6 +31,7 @@ import dataclasses
 import json
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any
+from uuid import uuid4
 
 from rampart.common.text import safe_float, safe_str, safe_str_list
 
@@ -45,16 +46,14 @@ if TYPE_CHECKING:
 class JsonFileReportSink:
     """Writes the test run report to a JSON file.
 
-    Each run produces a timestamped file:
-    ``<output_dir>/run_report_2026-03-19T21-30-00.json``
-    If that name already exists, the sink appends ``_1``, ``_2``, and so on.
+    Each run produces a file named ``run_report_<timestamp>_<uuid>.json``.
+    The UTC timestamp includes milliseconds; the UUID distinguishes runs
+    created in the same millisecond. Existing files are never overwritten.
 
     Args:
         output_dir (Path): Directory to write report files into.
             Created automatically if it does not exist.
     """
-
-    _MAX_FILENAME_ATTEMPTS: int = 1000
 
     def __init__(self, *, output_dir: Path) -> None:
         """Initialize with an output directory for report files."""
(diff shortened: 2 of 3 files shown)


rioyu123:

rampart/reporting/json_file.py
```diff
@@ -62,14 +65,41 @@ async def emit_async(self, *, report: TestRunReport) -> None:
 
         Args:
             report (TestRunReport): The aggregated test run results.
+
+        Raises:
+            FileExistsError: If no collision-free filename can be reserved.
         """
         self._output_dir.mkdir(parents=True, exist_ok=True)
 
         timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%S")
```
nina-msft: Feel free to also include milliseconds in the timestamp, to make it more concise and reduce the likelihood of collisions:

```suggestion
        timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%S-%f")[:-3]
```

nina-msft: If you update this, please also update docs that reference this filename :-)

- `rampart/reporting/json_file.py:48-50` — class docstring example.

- `docs/usage/results-and-reporting.md:90` — documented output example.

- `tests/unit/reporting/test_json_file.py:371-426` — exact filename expectations.

rampart/reporting/json_file.py
```diff
@@ -47,12 +47,15 @@ class JsonFileReportSink:
...
     Each run produces a timestamped file:
     ``<output_dir>/run_report_2026-03-19T21-30-00.json``
+    If that name already exists, the sink appends ``_1``, ``_2``, and so on.
 
     Args:
         output_dir (Path): Directory to write report files into.
             Created automatically if it does not exist.
     """
 
+    _MAX_FILENAME_ATTEMPTS: int = 1000
```
nina-msft: 1000 as a max value seems excessive for this change, lets reduce to 10? Unless you are using highly concurrent multi-process writers, in which case 100 would be safer.

rampart/reporting/json_file.py
```diff
@@ -62,14 +65,41 @@ async def emit_async(self, *, report: TestRunReport) -> None:
...
-        filepath = self._output_dir / f"run_report_{timestamp}.json"
-
         data = self._serialize_report(report)
-        filepath.write_text(json.dumps(data, indent=2, default=str))
+        self._write_report_file(
+            stem=f"run_report_{timestamp}",
+            content=json.dumps(data, indent=2, default=str),
+        )
+
+    def _write_report_file(self, *, stem: str, content: str) -> None:
```
nina-msft: Is it important to you for the file names to be ordered by which was generated first within the timestamp?

If not, maybe consider this approach which provides a more concise solution that still avoids overwrites (a readable millisecond timestamp plus a UUID):

```python
from uuid import uuid4

timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%S-%f")[:-3]
filepath = self._output_dir / f"run_report_{timestamp}_{uuid4().hex}.json"
with filepath.open("x", encoding="utf-8") as report_file:
    report_file.write(json.dumps(data, indent=2, default=str))
```

This removes the suffix loop, attempt limit, and helper method. `open("x")` still guarantees no existing file is overwritten. File names look like:

```
run_report_2026-08-27T13-33-17-840_a3f18c92654d4b75ad15687d383d951b.json
run_report_2026-08-27T13-33-17-840_647e96193ec34b19a54eaad43fd82151.json
```

Alternatively: Microseconds or `time.time_ns()` alone are shorter, but neither guarantees uniqueness across concurrent processes. Timestamp plus UUID is the clearest robust option.

nina-msft (dismissed):
Thanks for flagging this @rioyu123! Let me know what you think of the following comments :-)

rampart/reporting/json_file.py
rioyu123: That makes sense; there's no need to preserve ordering within a millisecond here. I've switched to a UTC millisecond timestamp plus `uuid4().hex` and removed the suffix loop, retry limit, and helper.

The file is still opened with `"x"`, so even a forced UUID collision can't replace an existing report. Serialization happens before opening the file. Tests cover repeated timestamps, the UTC argument, legacy files, forced collisions, and serialization failures; the filename docs are updated too.

Validation: all 1,033 unit tests passed (2 skipped), pre-commit passed, and the strict docs build passed on Linux.

rampart/reporting/json_file.py
```diff
@@ -62,14 +64,19 @@ async def emit_async(self, *, report: TestRunReport) -> None:
...
+            FileExistsError: If the generated filename already exists, or
+                ``output_dir`` exists and is not a directory.
         """
         self._output_dir.mkdir(parents=True, exist_ok=True)
 
-        timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%S")
-        filepath = self._output_dir / f"run_report_{timestamp}.json"
-
+        timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%S-%f")[:-3]
+        filepath = self._output_dir / f"run_report_{timestamp}_{uuid4().hex}.json"
```
nina-msft: Actually, could we add the UUID only when the millisecond filename collides? For example:

```python
timestamp = datetime.now(UTC).strftime("%Y-%m-%dT%H-%M-%S-%f")[:-3]
content = json.dumps(data, indent=2, default=str)
filepath = self._output_dir / f"run_report_{timestamp}.json"

try:
    report_file = filepath.open("x", encoding="utf-8")
except FileExistsError:
    filepath = self._output_dir / f"run_report_{timestamp}_{uuid4().hex}.json"
    report_file = filepath.open("x", encoding="utf-8")

with report_file:
    report_file.write(content)
```

This keeps the common filename concise while preserving atomic no-overwrite behavior.

docs/usage/results-and-reporting.md
```diff
@@ -87,7 +87,9 @@ from rampart.reporting import JsonFileReportSink
 sink = JsonFileReportSink(output_dir=Path(".report"))
 ```
 
-Output: `.report/run_report_2026-04-25T14-30-00.json`
+Output: `.report/run_report_2026-04-25T14-30-00-123_a3f18c92654d4b75ad15687d383d951b.json`
+
+The filename contains a UTC timestamp (millisecond precision) and a random UUID. Reports created in the same millisecond receive different filenames. An exact filename collision raises `FileExistsError` instead of overwriting an existing report. Reports written within the same millisecond have no defined filename order relative to each other.
```
nina-msft: If we implement behavior that I commented on below, the exact filename collision will just be avoided by adding the uuid at the end and we shouldn't raise FileExistsError :-)

rampart/reporting/json_file.py
rioyu123: Done. The sink now opens `run_report_<millisecond timestamp>.json` exclusively first and appends a UUID only if that path already exists. The fallback is also opened exclusively, so the common filename stays concise and same-millisecond reports cannot overwrite each other. Tests cover both paths.

docs/usage/results-and-reporting.md
rioyu123: Agreed. The guide now shows the concise millisecond filename and explains that a same-timestamp collision is handled by appending a UUID. It no longer says that an ordinary timestamp collision raises `FileExistsError`.

nina-msft (approved):
Thanks for your contribution @rioyu123 :-)

PR #185, opened 2026-09-04
[FEAT]: Canonical versioned trace/result serializer (WS0-05 PR 1)

nina-msft:
## WS0-05 — PR 1: Canonical versioned trace/result serializer

Establishes the neutral schema API and migration policy. Existing reporting and xdist consumers are not rewired in this PR.

### What this lands
- **`Result.to_dict()` / `Result.from_dict()`** — the canonical, unversioned body codec, backed by one cached Pydantic adapter over `Result`. The existing standard dataclasses remain the live models; there is no parallel body model or set of nested encode/decode functions. Encoding uses Python-mode adapter output with explicit enum/datetime serializers, followed by a canonical reader pass before returning the body. This extra pass preserves existing reader depth support without introducing a new depth cap; interpreter/parser recursion limits still apply.
- **Adapter-local policies** — a private `GetPydanticSchema` hook configures copies of generated core-schema nodes. Public fields retain plain dictionary/datetime annotations; no class-wide Pydantic configuration or trace validators are attached to the live dataclasses. Other Pydantic adapters retain their normal behavior. The canonical adapter explicitly supplies its datetime/Path resolution namespace; `types.py` is unchanged from the PR base.
- **Thin `ResultRecord` envelope** in `rampart/core/serialization.py`. `serialize_record(record=...) -> str` serializes an explicitly constructed `ResultRecord` to JSON text; `deserialize_record(data=...) -> ResultRecord` reads JSON text. `ResultRecord.to_dict()` / `from_dict()` remain available for dictionary conversion and projections. Every record carries the single root version `rampart.trace.v1`; unsupported versions fail closed through the decoder registry. Body dicts are fragments, not standalone durable records.
- **Optional attribution** — `pytest_nodeid` and `result_index`, omitted when not recorded. Stable identity fields and durable artifact descriptors remain deferred until their implementations exist.
- **Strict boundary validation** — all nine nested types, finite numbers, JSON-only free-form maps, closed enums, and passthrough harm-category strings. Missing fields use declared defaults without retaining original field presence; explicit null is accepted only for nullable fields. Payload IDs are required on read rather than generated. Python ISO timestamp formatting is preserved. Strings and mapping keys, including attribution, must contain Unicode scalar values: Python surrogate code points are rejected without lossy normalization, while valid JSON surrogate-pair escapes for characters such as emoji remain supported. Malformed JSON, unsupported values, recursion-limit failures, and adapter serialization `ValueError`s surface as `SchemaError`.
- **Trace-only payload restrictions** — binary formats and non-null artifacts fail before filesystem access; ordinary live binary payload construction and independent Pydantic validation are unchanged.
- **Nonmutating metadata hygiene** — `ResultRecord.to_dict()` filters only top-level transport bookkeeping from encoded output, including `_rampart_source_worker`. Body-only serialization and record decoding retain these keys. Re-encoding filters them from output without modifying the referenced result; nested user maps remain intact.
- **Generated open JSON Schema** at `schemas/trace.v1.schema.json`. `ResultRecord.json_schema()` composes the adapter-derived body and versioned envelope, with narrow customizations for trace payload restrictions and the Request invariant. Runtime class docstrings are omitted in favor of concise trace-specific descriptions. `scripts/generate_trace_schema.py --check` is the CI drift gate.
- **Explicit structural/semantic distinction** — passing JSON Schema validation is necessary but not sufficient for decoding. Integer notation, Unicode scalar strings, Python ISO datetime parseability, and finite representable numbers remain decoder requirements. These requirements and runtime nesting limits are documented rather than claimed as JSON Schema guarantees.
- **Migration and integration policy** at `docs/concepts/trace-schema.md`: additive optional fields require no bump; structural changes require a new major; absence means not recorded, not a semantic negative. Reserved future field names and the release-based support window are documented. Transport normalization must prepare a separate result before encoding; a binary text placeholder is a lossy transport view, not a durable replay artifact.

### Coverage
Hypothesis-generated body and record round-trips over nested JSON values, all nine nested types, enums, optional fields, and timestamps. Example-based coverage includes field exhaustiveness, version dispatch, attribution, additive tolerance, invalid JSON/types/cycles, independent-adapter isolation, schema semantics, and payload rejection before filesystem access. Additional regressions cover surrogate-containing text and mapping keys, valid Unicode/JSON surrogate-pair escapes, deep-nesting round-trips, and recursion/error boundaries with exception chaining. A prepared-copy boundary case demonstrates canonical reuse without modifying or relaxing the original result. `hypothesis` and `jsonschema` are development-only dependencies.

### Scope
Reporting projections, xdist/failure-attachment adoption, and stable identity remain separate follow-up implementations. Existing transport coercion and payload-display behavior require explicit preparation and compatibility work, not a mechanical serializer replacement. No public transport-preparation API or artifact representation is introduced here.


rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""Canonical, versioned trace/result serialization for RAMPART.
+
+This module owns the *single* full-fidelity ``Result`` <-> ``dict`` round-trip
+for the whole framework (design gate WS0-05, Decision D6). xdist transport,
```
nina-msft: remove ref to design doc

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+failure attachments, reporting projections, and future replay all serialize
+through here rather than maintaining parallel serializers.
+
+The canonical layer defines the supported *value domain* and nothing else. It
+does not apply transport hygiene — no ANSI stripping, no float normalization,
+no ``repr()``/``str()`` fallback, and no size capping. Those concerns wrap the
```
nina-msft: end sentence after "hygiene"

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+canonical output at the transport boundary (xdist). When a value falls outside
+the canonical domain the codec fails closed with a field path rather than
+coercing, so a durable trace never silently loses fidelity.
+
+Every serialized record carries a single root ``version`` field
+(:data:`TRACE_SCHEMA_VERSION`). Decoding dispatches on that version and fails
+closed on an unknown major. The record version is independent of the xdist
```
nina-msft: remove last sentence that references xdist

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+    Turn,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+EnumT = TypeVar("EnumT", bound=Enum)
+
+TRACE_SCHEMA_VERSION = "rampart.trace.v1"
+"""Single root schema version stamped on every serialized record."""
```
nina-msft: comments should be above the variables they reference not below

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+
+    This is the public surface: :meth:`to_dict` / :meth:`from_dict` are the one
+    round-trip every durable consumer uses. The ``result`` is referenced, not
+    copied. The collar fields (``identity``, ``pytest_nodeid``, ``result_index``)
+    are wire-only provenance stamped once at the producing boundary.
+
+    Args:
+        result (Result): The single-run verdict being serialized.
+        identity (dict[str, Any] | None): Stable test identity descriptor
+            (WS0-06). ``None`` until identity is wired at the producer.
```
nina-msft: remove ref to WS0-06

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+    case_id: str | None = None,
+    pytest_nodeid: str | None = None,
+    result_index: int = 0,
+) -> dict[str, Any]:
+    """Serialize a result to the canonical, versioned dict.
+
+    Args:
+        result (Result): The verdict to serialize.
+        identity (str | None): Stable identity value (WS0-06), if computed.
+        origin (str | None): How the identity was derived (marker vs. derived).
```
nina-msft: why is origin needed?

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+    return {
+        "kind": effect.kind,
+        "details": _encode_json(value=effect.details, path=f"{path}.details"),
+    }
+
+
+def _encode_payload(*, payload: Payload, path: str) -> dict[str, Any]:
+    """Encode a ``Payload``.
+
+    Binary payloads are persisted as content-addressed artifact descriptors by
```
nina-msft: remove this sentence since it references ws7  which hasn't yet been implemented

rampart/core/serialization.py
```diff
@@ -0,0 +1,913 @@
...
+
+    Returns:
+        dict[str, Any]: The encoded payload.
+
+    Raises:
+        SchemaError: If the payload uses a binary format.
+    """
+    if payload.format.is_binary:
+        msg = (
+            f"{path}: binary payload format {payload.format.value!r} requires the "
```
nina-msft: adjust to be generic unsupported in version message w/o ref to ws7

nina-msft: Changed both encode and decode errors to state generically that the binary format is unsupported in the current schema version.

rampart/core/serialization.py
nina-msft: Updated this to end after “transport hygiene” and removed the implementation-detail list.

rampart/core/serialization.py
nina-msft: Moved both explanatory comments above their constant declarations.

rampart/core/serialization.py
nina-msft: Removed the design-gate and work-item references from the module docstring.

rampart/core/serialization.py
nina-msft: Removed the WS0-06 references from both identity descriptions.

rampart/core/serialization.py
nina-msft: Removed the xdist version comparison from the module docstring.

rampart/core/serialization.py
nina-msft: Removed the planned artifact-descriptor and WS7 explanation. The docstring now describes only the current encode behavior.

rampart/core/serialization.py
nina-msft: It was not needed by serialization and would have committed this schema to identity-derivation details before that producer exists. I removed `origin` from the function, encoded identity descriptor, and tests.

rampart/core/result.py
```diff
@@ -157,7 +180,7 @@ class Result:
         default_factory=list[InjectionRecord],
     )
     population: PopulationRef | None = None
-    metadata: dict[str, Any] = field(default_factory=dict[str, Any])
+    metadata: JsonMapping = field(default_factory=dict[str, Any])
```
nina-msft: why change this to `JsonMapping`? will that mess with the api of result

rampart/core/result.py
```diff
@@ -110,6 +130,9 @@ class PopulationRef:
     threshold: float
 
 
+@with_config(
```
nina-msft: what is the functionality of this?

rampart/core/types.py
```diff
@@ -195,9 +263,9 @@ class ToolCall:
     """
 
     name: str
-    arguments: dict[str, Any] = field(default_factory=dict[str, Any])
+    arguments: JsonMapping = field(default_factory=dict[str, Any])
```
nina-msft: why update to JsonMapping and IsoDateTime in this file?

rampart/core/result.py
nina-msft: `@with_config(...)` attaches configuration that Pydantic uses when building `TypeAdapter(Result)`. It does not turn `Result` into a Pydantic dataclass or validate ordinary constructor calls/assignments.

- `strict=True` prevents coercions such as a string or boolean into an integer field.
- `revalidate_instances="always"` checks existing dataclass instances rather than trusting their types. This matters for `to_dict()`: a result or nested response may have been mutated after construction.
- `allow_inf_nan=False` rejects NaN/infinity in typed numeric fields. The separate `JsonMapping` validator covers values under `Any`.

The configuration propagates to the nested standard dataclasses while building this adapter, so they do not each need their own configuration. Like the annotated fields, this configuration is attached to the class: other Pydantic consumers of `Result` also see it, not just the canonical serializer.

rampart/core/types.py
nina-msft: The adapter recursively reads these dataclasses, so this attaches the two shared policies directly to the fields instead of writing separate ToolCall/Turn/Response codecs.

`JsonMapping` checks free-form arguments, details, and metadata recursively; `dict[str, Any]` alone does not prevent Pydantic from lossily serializing unsupported values.

`IsoDatetime` keeps the previous `datetime.isoformat()` / `datetime.fromisoformat()` behavior. Pydantic's defaults change UTC output from `+00:00` to `Z` and reject some Python-supported offsets on read. The alias preserves those round trips, including naive datetimes and subminute offsets. Its schema uses a string description rather than the narrower RFC 3339 `date-time` format.

Both are `Annotated` aliases: live values remain ordinary dictionaries and datetimes, and normal dataclass construction is unchanged. The tradeoff is that other Pydantic adapters inspecting these types also pick up these policies; they are not scoped to the canonical adapter. This placement reduces codec duplication, but adapter-specific hooks would isolate trace behavior better if we want the public field annotations left untouched.

rampart/core/result.py
nina-msft: `JsonMapping` is an alias for `Annotated[dict[str, Any], BeforeValidator(json_value)]`, not a new container. Callers still pass and receive ordinary dictionaries; the standard dataclass constructor and assignments do not run this validator. Static typing still sees `dict[str, Any]`.

The validator closes a gap in Pydantic's handling of `Any`: otherwise tuples can become lists and non-finite values can be normalized during serialization. It recursively rejects values outside the canonical JSON domain before that happens.

There is an API caveat: this is not completely invisible. Annotation introspection sees the added metadata, and other Pydantic consumers validating `Result` honor it too—even without our trace context. Ordinary dataclass use is unchanged, but Pydantic integration behavior is stricter. Attaching the policy here avoids a parallel model; it is an implementation choice, not a TypeAdapter requirement. Keeping the public annotations untouched would require moving this policy into adapter-specific customization.

rampart/core/result.py
```diff
@@ -194,6 +221,137 @@ def __repr__(self) -> str:
             f"summary={self.summary!r})"
         )
 
+    def to_dict(self) -> dict[str, Any]:
```
nina-msft, line 224: do we need these `to_dict` / `from_dict` functions?

nina-msft: Removed all four `to_dict()` / `from_dict()` methods in 34b13a2. `serialize_record()` and `deserialize_record()` now own conversion in the neutral serialization module, including the cached adapter and version dispatch. `Result` no longer carries serialization or schema-generation machinery; `ResultRecord.json_schema()` remains available. Structured consumers can parse the canonical JSON rather than needing another public conversion API. The v1 wire format and existing validation policies are unchanged.

.github/workflows/ci.yml
```diff
@@ -50,6 +50,9 @@ jobs:
       - name: ty
         run: uv run ty check
 
+      - name: Trace schema drift
+        run: uv run python scripts/generate_trace_schema.py --check
```
behnam-o: This drift check verifies that the generated schema matches the checked-in file, but it does not require a schema-version change. A breaking `Result` change can regenerate `trace.v1.schema.json` while leaving `TRACE_SCHEMA_VERSION` unchanged, and CI will pass with incompatible records still labeled `rampart.trace.v1`. Could model/schema changes require an explicit decision to either confirm v1 compatibility or introduce a new version?

rampart/core/serialization.py
```diff
@@ -0,0 +1,330 @@
...
+    Raises:
+        SchemaError: If the record cannot be represented as canonical JSON.
+    """
+    if not isinstance(record.result.metadata, Mapping):
+        msg = "result.metadata: expected a mapping"
+        raise SchemaError(msg)
+    metadata = {
+        key: value
+        for key, value in record.result.metadata.items()
+        if key not in _RESERVED_METADATA_KEYS
```
behnam-o: Could this pytest/xdist-specific filtering move to the follow-up consumer integration? No transport consumes this serializer in this PR, so that change would let the required bookkeeping and loss/truncation behavior be validated against the real consumer while keeping this layer focused on faithfully serializing the canonical `Result` domain.

docs/concepts/trace-schema.md
```diff
@@ -0,0 +1,237 @@
...
+  (`vN-1 → vN`);
+- migrating persisted data is an explicit operation; reading never rewrites an
+  artifact in place; and
+- encountering an unsupported major fails closed.
+
+## Reserved additive fields (named now, populated later)
+
+These record-level wire-only collar slots are reserved by name so they can be
+added without a major bump:
+`manifest_snapshot`, `evaluation_fingerprint`, `replay_provenance`,
```
behnam-o: Could these reserved names be removed until they have producers and consumers? The policy already permits additive optional fields without a version bump, so reserving fields that are neither emitted nor read provides no current compatibility benefit and may prematurely constrain their eventual design.

rampart/core/serialization.py
```diff
@@ -0,0 +1,330 @@
...
+    }
+    adapter = _result_adapter()
+    try:
+        validated = adapter.validate_python(
+            replace(record.result, metadata=metadata), strict=True
+        )
+        # JSON-mode dumping has a lower nesting limit than the reader.
+        body = adapter.dump_python(validated, mode="python", warnings="error")
+        _validate_body(body)
```
behnam-o: This makes the writer serialize in Python mode and then validate its own output through the reader, primarily to support custom deep-nesting behavior. Could we identify a realistic supported `Result` that requires this path? If the only case is synthetic nesting near interpreter limits, using the normal Pydantic JSON serialization path now and adding this workaround when a consumer demonstrates the need would keep the v1 foundation simpler.

rampart/core/serialization.py
```diff
@@ -0,0 +1,330 @@
...
+        """
+        if self.pytest_nodeid is not None and not isinstance(self.pytest_nodeid, str):
+            msg = "record.pytest_nodeid: expected a string or null"
+            raise SchemaError(msg)
+        if self.pytest_nodeid is not None:
+            try:
+                json_string(value=self.pytest_nodeid, path="record.pytest_nodeid")
+            except ValueError as exc:
+                raise SchemaError(str(exc)) from exc
+        if self.result_index is not None and type(self.result_index) is not int:
```
behnam-o: This type check still accepts negative `result_index` values even though the field is a within-node ordinal. The canonical boundary similarly accepts invalid `PopulationRef` relationships such as negative indices/sizes, `index >= size`, and thresholds outside `[0, 1]`. Could these semantic invariants be enforced here/nested validation and represented in the generated schema?

behnam-o (dismissed):
Overall, this looks pretty good to me. Please take a look at the inline comments. The schema-version verification comment is materially more important than the others: the current drift check can accept a regenerated, breaking v1 schema without requiring `TRACE_SCHEMA_VERSION` to change. The remaining comments are primarily about keeping this foundation scoped and deferring consumer-specific or speculative behavior.

.github/workflows/ci.yml
nina-msft: Agreed—the drift check alone did not enforce a compatibility decision. Added a separate CI gate in 4ffced8 with a checked-in `schemas/trace-compatibility.json` declaration tied to SHA-256 fingerprints of the models, codec policies, and published schemas. PR CI compares it with the actual target-base contract: changed inputs require an explicit same-major `compatible` decision and rationale, or a `new-major` decision with an adjacent version bump, retained historical schemas, and a migration-document reference. The first v1 introduction is explicitly declared `initial`. The schema generator also now selects its filename from `TRACE_SCHEMA_VERSION`, rather than always rewriting v1.

This enforces a recorded decision, not an automatic proof of compatibility: reviewers still assess the rationale, semantics, and required upcaster/migration API implementation. The documentation makes that boundary explicit.

rampart/core/serialization.py
nina-msft: Moved this out of the canonical codec in 2cb59be. Encoding and decoding now preserve supported metadata, including bookkeeping and loss/truncation markers; unsupported values fail validation regardless of the key name. The reserved-key list and filtered Result copy are removed, and the codec still leaves the source result untouched.

Shared, nonmutating consumer preparation now owns bookkeeping separation. Updated the reporting follow-up item #12820 and xdist follow-up item #10596 to require one shared policy, real-consumer coverage, preservation of nested user metadata/payload diagnostics, and retention or explicit relocation of loss/truncation evidence rather than silently discarding it. No consumer adoption or replacement serializer is introduced here.

docs/concepts/trace-schema.md
nina-msft: Removed the reserved-name list in 4008be9 and replaced it with general extension guidance: introduce fields when their producers, consumers, and absence semantics are defined; wire-only attribution belongs on ResultRecord and intrinsic data belongs on Result. Optional additions still use the existing compatibility policy, without promising compatibility for an unimplemented representation. Removed the remaining named speculative example and artifact-reservation wording as well. No runtime fields or wire schema changed.

rampart/core/serialization.py
nina-msft: Agreed—the motivating cases were synthetic deep-nesting inputs, not a demonstrated consumer requirement. Simplified the writer in 77ded00 to validate the live Result and use normal Pydantic JSON-mode serialization, without feeding its output back through the reader. Removed the Python-mode enum hook as well; native JSON enum encoding handles it.

Unicode-scalar validation now runs on typed strings during adapter validation, so removing the readback pass does not weaken that guarantee. Python ISO datetime formatting and SchemaError wrapping remain. The docs explicitly state that reader/writer nesting limits can differ: a deeply nested external record may decode successfully but fail to re-encode. Coverage now reflects that limitation rather than promising depth parity.

rampart/core/serialization.py
nina-msft: Added these invariants in 323223a. ResultRecord rejects a negative result_index; the canonical adapter enforces PopulationRef size >= 1, 0 <= index < size, and threshold in [0, 1] on both encoding and decoding. The policies remain adapter-local, so live PopulationRef construction and independent Pydantic adapters are unchanged.

The generated schema includes minimum/maximum constraints for the scalar bounds. Standard JSON Schema cannot compare index with the sibling size value, so that relationship is enforced at runtime and explicitly documented in the generated schema and policy page. Coverage includes invalid and boundary values, live-result changes after wrapping, both conversion directions, schema constraints, and adapter isolation. The compatibility declaration records these as refinements of the initial unreleased contract, not a claim that tightening an already-published v1 would be compatible.

behnam-o (approved):
The latest revision addresses my review comments correctly. The compatibility gate now requires an explicit schema-version decision, and the scope, serialization path, metadata behavior, and invariants look good. Focused tests and static checks pass.

